PR #18, opened 2016-06-23
SDK-101 building and deploying module before run

AdamGrzybkowski:
Workflow: 
-tempProject folder with modules from watched projects is created
-pom.xml of each module is extended by parent
-pom.xml of tempProject is created
-"mvn clean instal" in tempProject
-deploying modules to server

I will try to add some  integration test

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
@@ -105,6 +105,14 @@ public class Run extends AbstractTask {
 		}
 	}
 
+	/**
+	 * Creates temporary Maven reactor project to build dependencies in the correct order
+	 *
+	 * @param server
+	 * @return
+	 * @throws MojoFailureException
+	 * @throws MojoExecutionException
+     */
 	private File createTempReactorProject(Server server) throws MojoFailureException, MojoExecutionException {
 		File tempFolder = new File(server.getServerDirectory(), "temp-project");
 		if (!tempFolder.exists()) {
@@ -135,6 +143,11 @@ public class Run extends AbstractTask {
 		return tempFolder;
 	}
 
+	/**
+	 * Deletes temporary Maven reactor project
+	 *
+	 * @param tempFolder
+     */
 	private void deleteTempReactorProject(File tempFolder) {
 		try {
 			FileUtils.deleteDirectory(tempFolder);
@@ -143,6 +156,11 @@ public class Run extends AbstractTask {
 		}
 	}
 
+	/**
+	 * Creates Model to generate pom.xml for temporary project
+	 *
+	 * @return
+     */
 	private Model createModel(){
 		Model model = new Model();
 		model.setArtifactId(String.format("openmrs-sdk-server-%s", serverId));
@@ -154,6 +172,14 @@ public class Run extends AbstractTask {
 		return model;
 	}
 
+	/**
+	 * Deploy all watched modules to server
+	 *
+	 * @param server
+	 * @throws MojoFailureException
+	 * @throws MojoExecutionException
+	 * @throws MavenInvocationException
+     */
 	private void deployWatchedModules(Server server) throws MojoFailureException, MojoExecutionException, MavenInvocationException {
 		Set<Project> watchedProject = server.getWatchedProjects();
 		for (Project module: watchedProject) {
@@ -162,6 +188,11 @@ public class Run extends AbstractTask {
 		}
 	}
 
+	/**
+	 * Run "mvn clean install -DskipTests" command in the given directory
+	 * @param directory
+	 * @throws MojoFailureException
+     */
 	private void cleanInstallServerProject(String directory) throws MojoFailureException {
 		String maven = "mvn";
 		if (System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("windows")) {
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
@@ -1,18 +1,28 @@
 package org.openmrs.maven.plugins;
 
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.maven.model.Model;
+import org.apache.maven.model.io.xpp3.MavenXpp3Writer;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.shared.invoker.MavenInvocationException;
 import org.openmrs.maven.plugins.model.Artifact;
 import org.openmrs.maven.plugins.model.Server;
 import org.openmrs.maven.plugins.utility.Project;
 import org.openmrs.maven.plugins.utility.SDKConstants;
 
 import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
+import java.io.Writer;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.Set;
 
 import static org.twdata.maven.mojoexecutor.MojoExecutor.*;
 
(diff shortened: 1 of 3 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
 			}
 		}
 
+		try {
+			File tempFolder = createTempReactorProject(server);
+			cleanInstallModule(tempFolder.getPath());
+			deleteTempReactorProject(tempFolder);
+			deployWatchedModules(server);
+		} catch (MavenInvocationException e) {
+			throw new MojoFailureException(e.getMessage());
```
rkorytkowski: You should always try to preserve a stacktrace. Use new MojoFailureException("Failed to build watched modules", e);

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
 			}
 		}
 
+		try {
+			File tempFolder = createTempReactorProject(server);
```
rkorytkowski: I would start from putting an if(project.hasWatchedModules()) to make it clear for someone reading the code to ignore the whole try block if the list is empty.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+			} catch (IOException e) {
+				e.printStackTrace();
+			}
+		}
+
+		try {
+			Writer writer = new FileWriter(new File(tempFolder, "pom.xml"));
+			new MavenXpp3Writer().write(writer, tempModel);
+		} catch (IOException e) {
+			e.printStackTrace();
```
rkorytkowski: Do not eat the exception.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+				Parent parent = new Parent();
+				parent.setArtifactId("temp-project");
+				parent.setGroupId("org.temp");
+				parent.setVersion("1.0");
+				model.setParent(parent);
+				Writer writer  = new FileWriter(new File(moduleFile, "pom.xml"));
+				new MavenXpp3Writer().write(writer, model);
+				tempModel.addModule(module.getArtifactId());
+			} catch (IOException e) {
+				e.printStackTrace();
```
rkorytkowski: Do not eat the exception, rethrow as runtime instead.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+			File srcDir = new File(module.getPath());
+			File moduleFile = new File(tempFolder, module.getArtifactId());
+			try {
+				FileUtils.copyDirectory(srcDir, moduleFile);
+				Model model = Project.loadProject(moduleFile).getModel();
+				Parent parent = new Parent();
+				parent.setArtifactId("temp-project");
+				parent.setGroupId("org.temp");
+				parent.setVersion("1.0");
+				model.setParent(parent);
```
rkorytkowski: We should not be setting parents. It should be enough to add them as modules to the temp-project.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+				Writer writer  = new FileWriter(new File(moduleFile, "pom.xml"));
+				new MavenXpp3Writer().write(writer, model);
+				tempModel.addModule(module.getArtifactId());
```
rkorytkowski: I would have thought you need to add a path here, not just the artifactId...

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+		if (!tempFolder.exists()) {
+			tempFolder.mkdir();
+		}
+		Model tempModel = createModel();
+		Set<Project> watchedModules = server.getWatchedProjects();
+		for(Project module: watchedModules){
+			File srcDir = new File(module.getPath());
+			File moduleFile = new File(tempFolder, module.getArtifactId());
+			try {
+				FileUtils.copyDirectory(srcDir, moduleFile);
```
rkorytkowski: I understand now... You copy the source to temp-project thus you can use artifactId in line 120... Could you please try simply adding a path instead to the original project? I think some time ago I make it work with relative paths, but maybe absolute paths are working too...

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+		try {
+			FileUtils.deleteDirectory(tempFolder);
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
+	}
+
+	private Model createModel(){
+		Model model = new Model();
+		model.setArtifactId("temp-project");
```
rkorytkowski: It should be unique across maven repo. Please change it to org.openmrs:openmrs-sdk-server-{serverId}:1.0.0-SNAPSHOT.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+		model.setVersion("1.0");
+		model.setGroupId("org.temp");
```
rkorytkowski: It should be unique across maven repo. Please change it to org.openmrs:openmrs-sdk-server-{serverId}:1.0.0-SNAPSHOT.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+
+	private void deployWatchedModules(Server server) throws MojoFailureException, MojoExecutionException, MavenInvocationException {
+		Set<Project> watchedProject = server.getWatchedProjects();
+		for (Project module: watchedProject) {
+			Project project = Project.loadProject(new File(module.getPath()));
+			new Deploy(this).deployModule(project.getGroupId(), project.getArtifactId(), project.getVersion(), server);
+		}
+	}
+
+	private void cleanInstallModule(String directory) throws MojoFailureException {
```
rkorytkowski: Rename to cleanInstallServerProject

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+		ProcessBuilder processBuilder = new ProcessBuilder(commands);
+		processBuilder.redirectErrorStream(true);
+		processBuilder.redirectOutput(ProcessBuilder.Redirect.INHERIT);
+		processBuilder.redirectInput(ProcessBuilder.Redirect.INHERIT);
+		processBuilder.directory(new File(directory));
+		try {
+			final Process process = processBuilder.start();
+			process.waitFor();
+		} catch (IOException e) {
+			throw new MojoFailureException("Failed to build module", e);
```
rkorytkowski: Failed to build server project

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+		} catch (InterruptedException e) {
+			throw new MojoFailureException("Interrupted waiting", e);
```
rkorytkowski: Failed to build server project

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -73,13 +81,113 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
 			}
 		}
 
+		try {
+			File tempFolder = createTempReactorProject(server);
+			cleanInstallModule(tempFolder.getPath());
+			deleteTempReactorProject(tempFolder);
```
rkorytkowski: Probably this should be executed in a finally block to make sure we always clean up properly.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
AdamGrzybkowski: adding absolute path to the pom does not work, maven is adding this path to the server path so it cannot find modules.

rkorytkowski: Okay, how about we create a link instead? I would like to avoid copying, because it unnecessarily uses up space and takes too much time. See https://docs.oracle.com/javase/tutorial/essential/io/links.html

rkorytkowski:
Looks good, could you please update it to the latest from repo?

AdamGrzybkowski:
@rkorytkowski done

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -77,13 +87,140 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
+		if (server.hasWatchedProjects()) {
+			File tempFolder = createTempReactorProject(server);
+			cleanInstallServerProject(tempFolder.getPath());
+			deleteTempReactorProject(tempFolder);
+			try {
+				deployWatchedModules(server);
+			} catch (MavenInvocationException e) {
+				throw new MojoFailureException("Failed to deploy watched modules", e);
+			}
+		}
```
rkorytkowski: Put the whole if in a separate method, which you will be able to test.

PR #19, opened 2016-06-24
SDK-96 - Added JDK choose option

tmarzeion:
This PR is clone of https://github.com/openmrs/openmrs-contrib-sdk-maven-plugin/pull/39 and it will be ready to merge when fixes suggested by @rkorytkowski are applied

### DIFF
--- a/integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
+++ b/integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
@@ -4,10 +4,19 @@ import org.junit.Ignore;
 import org.junit.Test;
 import org.openmrs.maven.plugins.model.DistroProperties;
 import org.openmrs.maven.plugins.model.Server;
+import org.openmrs.maven.plugins.utility.SDKConstants;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
 import java.util.UUID;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.*;
+import static org.hamcrest.Matchers.*;
 import static org.openmrs.maven.plugins.SdkMatchers.serverHasVersion;
 
 public class SetupIntegrationTest extends AbstractSdkIntegrationTest {
@@ -89,5 +98,43 @@ public class SetupIntegrationTest extends AbstractSdkIntegrationTest {
         assertFilePresent(serverId+"/openmrs-1.11.5.war");
         assertFilePresent(serverId+"/modules");
         assertModulesInstalled(serverId, "owa-1.4.omod", "uicommons-1.7.omod", "uiframework-3.6.omod");
+    }
+
+    @Test
+    public void setup_shouldInstallServerWithGivenJdkAndCreateSdkPropertiesFileIfNotExist() throws Exception{
+        String serverId = UUID.randomUUID().toString();
+        addTaskParam("serverId", serverId);
+
+        executeTask("setup");
+        assertSuccess();
+        assertServerInstalled(serverId);
+
+        assertFilePresent(serverId + File.separator + "openmrs-server.properties");
+
+        File serverPropertiesFile = new File(testDirectory.getAbsolutePath() + File.separator + serverId, "openmrs-server.properties");
+        String javaHomeServerProperty = readValueFromPropertyKey(serverPropertiesFile, "java.home");
+        String defaultJavaHome = System.getProperty("java.home");
+        assertThat(defaultJavaHome, is(javaHomeServerProperty));
+        assertFilePresent("/sdk.properties");
+    }
+
+    @Test
+    public void setup_shouldInstallServerWithGivenJdkInBatchMode() throws Exception{
+        String serverId = UUID.randomUUID().toString();
+        addTaskParam("serverId", serverId);
+        addTaskParam("java.home", System.getProperty("java.home"));
+
+        executeTask("setup");
+        assertSuccess();
+        assertServerInstalled(serverId);
+    }
+
+    private String readValueFromPropertyKey(File propertiesFile, String key) throws Exception {
+
+        InputStream in = new FileInputStream(propertiesFile);
+        Properties sdkProperties = new Properties();
+        sdkProperties.load(in);
+
+        return sdkProperties.getProperty(key);
     }
 }
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.Map;
 
 import static org.twdata.maven.mojoexecutor.MojoExecutor.*;
 
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
@@ -81,6 +81,13 @@ public class Setup extends AbstractTask {
     private String dbSql;
 
     /**
+     * Path to JDK Version
+     *
+     * @parameter expression="${java.home}"
+     */
+    private String jdk;
+
+    /**
      * Path to installation.properties
      *
      * @parameter expression="${file}"
@@ -179,6 +186,18 @@ public class Setup extends AbstractTask {
             } else {
                 moduleInstaller.installModule(SDKConstants.H2_ARTIFACT, server.getServerDirectory().getPath());
                 wizard.showMessage("The specified database "+server.getDbName()+" does not exist and it will be created for you.");
+            }
+        }
+
+        if (server.getJavaHome() == null || server.getJavaHome().equals(System.getProperty("java.home"))) {
+            wizard.promptForJdkPath(server);
+            if (server.getJavaHome() == null) {
+                server.setJavaHome(System.getProperty("java.home"));
+            }
+        }
+        else {
+            if (!wizard.isThereJdkUnderPath(server.getJavaHome())) {
+                throw new IllegalArgumentException("Given -Djdk property is invalid");
             }
         }
 
@@ -341,6 +360,7 @@ public class Setup extends AbstractTask {
                 .setDbUser(dbUser)
                 .setDbPassword(dbPassword)
                 .setInteractiveMode(interactiveMode)
+                .setJavaHome(jdk)
                 .build();
         wizard.promptForNewServerIfMissing(server);
 
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
@@ -25,6 +25,7 @@ public class Server {
     public static final String PROPERTY_DB_PASS = "connection.password";
     public static final String PROPERTY_DB_URI = "connection.url";
     public static final String PROPERTY_VERSION = "openmrs.version";
+    public static final String PROPERTY_JDK = "java.home";
     public static final String PROPERTY_PLATFORM = "openmrs.platform.version";
     public static final String PROPERTY_DB_NAME = "database_name";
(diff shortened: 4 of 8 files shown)


tmarzeion:
Changes suggested by @rkorytkowski are applied, but changes still needs to be covered by tests

tmarzeion:
@rkorytkowski Tests done

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -295,6 +298,10 @@ private void runInFork(Server server) throws MojoExecutionException, MojoFailure
 		processBuilder.redirectErrorStream(true);
 		processBuilder.redirectOutput(ProcessBuilder.Redirect.INHERIT);
 		processBuilder.redirectInput(ProcessBuilder.Redirect.INHERIT);
+		Map<String, String> env = processBuilder.environment();
+		String jdk = server.getJdkPath();
+		env.put("JAVA_HOME", jdk);
+		System.out.println("Using JAVA_HOME: " + jdk + "\n");
```
rkorytkowski: 292,293 above do it already. Please remove.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Run.java
```diff
@@ -270,6 +271,8 @@ private void runInFork(Server server) throws MojoExecutionException, MojoFailure
 		}
 
 		System.out.println("\nForking a new process... (use -Dfork=false to prevent forking)\n");
+		System.setProperty("MAVEN_OPTS", mavenOpts);
+		System.out.println("\n" + "Using MAVEN_OPTS: " + mavenOpts + "\n");
```
rkorytkowski: Please remove since lines 292,293 do it already. Also the output is done in RunTomcat.

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
```diff
@@ -508,6 +514,14 @@ public void removeFromValueList(String key, String artifactId) {
         }
     }
 
+    public String getJdkPath() {
```
rkorytkowski: Duplicates Server.getJavaHome(), please use that instead.

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
```diff
@@ -508,6 +514,14 @@ public void removeFromValueList(String key, String artifactId) {
...
+        return getParam(PROPERTY_JDK);
+    }
+
+    public void setJdkPath(String path) {
```
rkorytkowski: Rename to setJavaHome

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -270,6 +277,137 @@ public String promptForPlatformVersion(List<String> versions) {
...
+
+        if (!interactiveMode) {
+            return paths.get(0);
+        }
+
+        String path = promptForMissingValueWithOptions(SDKConstants.OPENMRS_SDK_JDK_OPTION,
+                server.getJdkPath(), "path", paths, SDKConstants.OPENMRS_SDK_JDK_CUSTOM, null);
+
+        if (path.equals(paths.get(0))) {
+            server.setJdkPath(System.getProperty("java.home"));
```
rkorytkowski: Do nothing in such a case so that the actual java.home system property is used and not its value during setup.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -270,6 +277,137 @@ public String promptForPlatformVersion(List<String> versions) {
...
 
+    @Override
+    public String promptForJdkPath(Server server) {
+
+        List<String> paths = new ArrayList<>();
+        paths.add("JAVA_HOME (currently: " + System.getProperty("java.home") + ")");
+        try {
+            paths.addAll(getJdkPathsFromProperties(getSdkProperties()));
+        } catch (IOException e) {
+            e.printStackTrace();
```
rkorytkowski: Never eat exceptions. Rethrow this one as IllegalArgumentException.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -270,6 +277,137 @@ public String promptForPlatformVersion(List<String> versions) {
...
+        if (!isThereJdkUnderPath(path)) {
+                System.out.println(SDKConstants.OPENMRS_SDK_JDK_CUSTOM_INVALID);
+                server.setJdkPath(promptForJdkPath(server));
+        }
+        else {
+            if (!isPathLineDuplicated(paths, path)) {
+                try {
+                    addJdkPathToSdkProperties(path);
+                } catch (IOException e) {
+                    e.printStackTrace();
```
rkorytkowski: Never eat exceptions. Rethrow as IllegalStateException

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -270,6 +277,137 @@ public String promptForPlatformVersion(List<String> versions) {
...
+                }
+            }
+            return path;
+        }
+        return null;
+    }
+
+    private void addJdkPathToSdkProperties(String path) throws IOException {
```
rkorytkowski: It's better to catch IOException and rethrow as IllegalStateException (RuntimeException) inside this method as there's nothing one can do when calling it.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -270,6 +277,137 @@ public String promptForPlatformVersion(List<String> versions) {
...
+    private String convertListToPropertyString(List<String> list) {
+        String result = "";
+        for (String str : list) {
+            result += str + ", ";
+        }
+        return result;
+    }
+
+    public Properties getSdkProperties() throws IOException {
+        File sdkPropertiesFile = new File(Server.getServersPathFile(), SDKConstants.OPENMRS_SDK_PROPERTIES);
```
rkorytkowski: Same here. No point in throwing IOException. You can catch it and rethrow as IllegalStateException.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -270,6 +277,137 @@ public String promptForPlatformVersion(List<String> versions) {
...
+            sdkPropertiesFile.createNewFile();
+        }
+
+        InputStream in = new FileInputStream(sdkPropertiesFile);
+        Properties sdkProperties = new Properties();
+        sdkProperties.load(in);
+        return sdkProperties;
+    }
+
+    public ArrayList<String> getJdkPathsFromProperties(Properties sdkProperties) {
```
rkorytkowski: Never return actual collection implementation ArrayList. Use the List inteface instead.

rkorytkowski: Wouldn't it be cleaner to just have getJdkPaths() instead of getJdkProperties and getJdkPathsFromProperties?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -270,6 +277,137 @@ public String promptForPlatformVersion(List<String> versions) {
...
+        List<String> result = new ArrayList<>();
+
+        if (!interactiveMode) {
+            return new ArrayList<>();
+        }
+
+        String jdkHomeProperty = sdkProperties.getProperty(SDKConstants.OPENMRS_SDK_PROPERTIES_JDK_HOME);
+        if (jdkHomeProperty != null) {
+            result.addAll(Arrays.asList(jdkHomeProperty.split("\\s*,\\s*")));
```
rkorytkowski: What does "\\s*,\\s*" do? Wouldn't it be easier to just use "," ?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -270,6 +277,137 @@ public String promptForPlatformVersion(List<String> versions) {
...
+
+            for (int i = 0; i < result.size(); i++) {
+                if (!isThereJdkUnderPath(result.get(i))) {
+                    result.remove(i);
+                    i--;
+                }
+            }
+
+            return (ArrayList<String>) result;
```
rkorytkowski: How about we save the properties to the file before returning?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -63,6 +63,13 @@
     private String dbPassword;
 
     /**
+     * Path to JDK Version
+     *
+     * @parameter expression="${jdk}"
```
rkorytkowski: let's actually make it java.home

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
```diff
@@ -90,4 +90,15 @@ public void setup_shouldInstallServerFromDistroPropertiesDir() throws Exception{
         assertFilePresent(serverId+"/modules");
         assertModulesInstalled(serverId, "owa-1.4.omod", "uicommons-1.7.omod", "uiframework-3.6.omod");
     }
+
+    @Test
+    public void setup_shouldInstallServerWithGivenJdkAndCreateSdkPropertiesFileIfNotExist() throws Exception{
+        String serverId = UUID.randomUUID().toString();
+        addTaskParam("serverId", serverId);
```
rkorytkowski: Please add a test which does addTaskParam("java.home", System.getProperty("java.home")); as well.

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
```diff
@@ -90,4 +90,15 @@ public void setup_shouldInstallServerFromDistroPropertiesDir() throws Exception{
...
+
+        executeTask("setup");
+        assertSuccess();
+        assertServerInstalled(serverId);
```
rkorytkowski: You should actually check here that java.home was set or wasn't set in openmrs-server.properties...

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
```diff
@@ -90,4 +90,15 @@ public void setup_shouldInstallServerFromDistroPropertiesDir() throws Exception{
...
+        assertFilePresent("/sdk.properties");
```
rkorytkowski: You should actually check if the properties were properly saved in the file and not only that file is present.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
tmarzeion: I just wanted to make sure that whitespaces won't mess with JDK paths in sdk.properties.
See this for details: http://stackoverflow.com/a/7488676

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
tmarzeion: I won't get rid of getSdkProperties() since its preventing code repeat, but I've changed getJdkPathsFromProperties(Properties sdkProperties) to getJdkPaths() (im getting properties from getSdkProperties() method inside getJdkPaths(), so i don't need to pass it through argument anymore).

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
tmarzeion: Its tough one since there its difficult to mock methods. Ill handle that later.
Except this one, every thing You asked for are completed.

PR #20, opened 2016-06-27
SDK-86 Install Open Web App

PawelGutkowski:
This PR introduces changes:

1. add `Bintray` class to interact with Bintray's Rest API and `OpenmrsBintray` to easily obtain packages from openmrs repositories on Bintray
2. add `Deploy OpenMRS OWA` option to `deploy` task  in interactive and batch mode (eg parameter `-Dowa=conceptdictionary:1.0.0-beta.6`)
3. store OWAs installed by user in server properties
4. create appropriate integration tests

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java
@@ -6,6 +6,9 @@ import org.apache.maven.plugin.BuildPluginManager;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.project.MavenProject;
+import org.openmrs.maven.plugins.bintray.BintrayId;
+import org.openmrs.maven.plugins.bintray.BintrayPackage;
+import org.openmrs.maven.plugins.bintray.OpenmrsBintray;
 import org.openmrs.maven.plugins.model.Artifact;
 import org.openmrs.maven.plugins.model.DistroProperties;
 import org.openmrs.maven.plugins.model.Server;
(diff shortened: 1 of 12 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java
```diff
@@ -151,13 +160,51 @@ private void runInteractiveMode(Server server, ServerUpgrader upgrader) throws M
...
+            for(BintrayId id : bintray.getAvailableOWA()){
+                owas.add(id.getName());
+            }
+            name = wizard.promptForMissingValueWithOptions("Which OWA would you like to deploy?%s", name, "", owas, true);
+        }
+        if(version == null){
+            List<String> versions = bintray.getOwaMetadata(name).getVersions();
+            version = wizard.promptForMissingValueWithOptions("Which version would you like to deploy?%s", version, "", versions, true);
+        }
+        wizard.showMessage("Please remember that to run OWA you need to have openmrs-owa-module installed on the server");
```
rkorytkowski: Could we actually check if it is installed here and if not install the latest version?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java
```diff
@@ -151,13 +160,51 @@ private void runInteractiveMode(Server server, ServerUpgrader upgrader) throws M
...
+        File owaDir = new File(server.getServerDirectory(), "owa");
+        if(!owaDir.exists()){
+            //couldn't find place where owa dir is saved yet
+            boolean useDefaultDir = wizard.promptYesNo(String.format(
+                    "\nThere is no default directory '%s' on server %s, would you like to create it? (if not, you will be asked for path to custom directory)",
```
rkorytkowski: Could you please explain this to me? Why would one want create a custom directory?

maven-plugin/src/main/java/org/openmrs/maven/plugins/bintray/Bintray.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.commons.httpclient.HttpClient;
+import org.apache.commons.httpclient.methods.GetMethod;
+import org.apache.commons.io.FileUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.util.List;
+
+public class Bintray {
```
rkorytkowski: I like that you created a wrapper!

maven-plugin/src/main/java/org/openmrs/maven/plugins/bintray/Bintray.java
```diff
@@ -0,0 +1,118 @@
...
+    public static List<BintrayId> getAvailablePackages(String owner, String repo){
```
rkorytkowski: Any reason those are static? Statics are harder to test/mock and can't be overriden when extending.

rkorytkowski:
Please address my comments in a separate PR.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java
PawelGutkowski: There is option in OWA module to use custom app directory, so I thought we should support that possibility.

maven-plugin/src/main/java/org/openmrs/maven/plugins/bintray/Bintray.java
PawelGutkowski: I thought that would make sense, as there is no state preserved here, but factors You listed seem more important. I will fix this and adress Your other comments in follow-up PR, thanks!

PR #21, opened 2016-06-27
SDK-102 openmrs-sdk:pull added

AdamGrzybkowski:
So, I'm not really sure if I did it in the proper way, but when catching exception instead of throwing one I'm setting the failureReason which is then printed to the user. 
I didn't throw for example RuntimeException, because I think it's better to try pull changes for every watched module and thet print which one were updated successfully and which one wasn't.

Another thing is when calling git.stashApply there is no way to check the stash message if there were any conflicts(like you can when doing pull rebase) so I'm reverting all changes in catch block.

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
@@ -1,6 +1,7 @@
 package org.openmrs.maven.plugins;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Model;
 import org.apache.maven.plugin.BuildPluginManager;
@@ -16,19 +17,22 @@ import org.eclipse.jgit.api.RebaseCommand;
 import org.eclipse.jgit.api.RemoteAddCommand;
 import org.eclipse.jgit.api.RemoteRemoveCommand;
 import org.eclipse.jgit.api.ResetCommand;
-import org.eclipse.jgit.api.errors.CheckoutConflictException;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.api.errors.StashApplyFailureException;
 import org.eclipse.jgit.internal.storage.file.FileRepository;
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.revwalk.RevCommit;
+import org.eclipse.jgit.transport.RemoteConfig;
 import org.eclipse.jgit.transport.URIish;
 import org.openmrs.maven.plugins.model.Server;
+import org.openmrs.maven.plugins.utility.CompositeException;
 import org.openmrs.maven.plugins.utility.Project;
 
 import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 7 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
```diff
@@ -0,0 +1,293 @@
...
+        try {
+            localRepo = new FileRepository(new File(path, ".git").getAbsolutePath());
+            git = new Git(localRepo);
+
+            previousBranch = localRepo.getBranch();
+
+            checkoutAndCreateNewBranch(git, newBranch);
+            newBranchFull = localRepo.getFullBranch();
+            RevCommit stash = git.stashCreate().call();
+            addRemoteTempRepo(git, path, newBranch);
```
rkorytkowski: It's better not to add a temp repo. Just pass uri to pullFromRemote.

AdamGrzybkowski: I agree, but it didn't work even the Jgit API says it should. Jgit is looking for this URL in remote repos and if not found it throws en exception. But maybe im missing something

AdamGrzybkowski: To be more specific, if the user didn't add remote to the upstream with the url from pom.xml JGit won't execute pull command

rkorytkowski: Okay, if it is the case then let's simply set/add remote 'upstream' not temporarily, but persistently.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
```diff
@@ -0,0 +1,293 @@
...
+                if (pullLatestUpstream(project.getPath())) {
+                    wizard.showMessage(String.format(PULL_SUCCESS_MESSAGE, project.getArtifactId()));
+                } else {
+                    getLog().error(String.format(PULL_ERROR_MESSAGE, project.getArtifactId(), failureReason));
+                    getLog().error(UPDATE_MANUALLY_MESSAGE);
+                    pullingError = true;
+                }
+            }
+            if(pullingError){
+                throw new MojoExecutionException("Some modules could not be updated automatically, please see error messages above");
```
rkorytkowski: I'd rather we list all modules that were updated and all that were not updated here.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
```diff
@@ -0,0 +1,293 @@
...
+
+        if(server.hasWatchedProjects()){
+            boolean pullingError = false;
+            Set<Project> watchedProjects = server.getWatchedProjects();
+            for(Project project: watchedProjects) {
+                wizard.showMessage(String.format(UPDATING_MODULE_MESSAGE, project.getArtifactId()));
+                if (pullLatestUpstream(project.getPath())) {
+                    wizard.showMessage(String.format(PULL_SUCCESS_MESSAGE, project.getArtifactId()));
+                } else {
+                    getLog().error(String.format(PULL_ERROR_MESSAGE, project.getArtifactId(), failureReason));
```
rkorytkowski: Let's introduce wizard.showError() instead of using log directly.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
```diff
@@ -0,0 +1,293 @@
...
+        return true;
+    }
+
+    /**
+     * Delete temporary remote
+     * @param git
+     * @param branchName
+     * @throws GitAPIException
+     */
+    private void deleteRemoteTempRepo(Git git, String branchName) throws GitAPIException {
```
rkorytkowski: Should not be needed. See my comment above about adding 'upstream' permanently.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
```diff
@@ -0,0 +1,293 @@
...
+                    failureReason = CONFLICT_MESSAGE;
+                    return false;
+                }
+            } else {
+                checkoutBranch(git, previousBranch);
+                mergeWithNewBranch(git, newBranchFull);
+                return true;
+            }
+
+        } catch (IOException e) {
```
rkorytkowski: Please put try catch on methods that actually fail with that exception and not around the whole block. In other words try catch should be as small as possible.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
```diff
@@ -0,0 +1,293 @@
...
+
+            checkoutAndCreateNewBranch(git, newBranch);
+            newBranchFull = localRepo.getFullBranch();
+            RevCommit stash = git.stashCreate().call();
+            addRemoteTempRepo(git, path, newBranch);
+            PullResult pullResult = pullFromRemote(git, newBranch);
+            if(!pullResult.isSuccessful()){
+                rebaseAbort(git);
+                checkoutBranch(git, previousBranch);
+                failureReason = CONFLICT_MESSAGE;
```
rkorytkowski: It's better to throw a checked exception with the given message instead of having failureReason.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
```diff
@@ -0,0 +1,293 @@
...
+                return false;
```
rkorytkowski: The method should be void. Catch exception if anything goes wrong.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
```diff
@@ -0,0 +1,293 @@
...
+                try {
+                    git.stashApply().setStashRef(stash.getName()).call();
+                    checkoutBranch(git, previousBranch);
+                    mergeWithNewBranch(git, newBranchFull);
+                    return true;
+                } catch (StashApplyFailureException e) {
+                    git.reset().setMode(ResetCommand.ResetType.HARD).setRef(previousBranch).call();
+                    checkoutBranch(git, previousBranch);
+                    git.stashApply().setStashRef(stash.getName()).call();
+                    failureReason = CONFLICT_MESSAGE;
```
rkorytkowski: It's better to throw a checked exception with your message and also passing the original exception.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
AdamGrzybkowski: Okay, I thought it's better to clean up after but I can do this that way :)

rkorytkowski: I just want to use this as an opportunity to introduce remote 'upstream' pointing to original project as a convention.

AdamGrzybkowski: I totally understand that, but what if there is an upstream remote but with a different url then specified in pom.xml ?

rkorytkowski: We want to overwrite that. Just display a message to the user that it will be corrected.

rkorytkowski:
Please add some tests once. It will be easier once SDK-103 is done.

PR #23, opened 2016-06-29
SDK-105 Import initial mysql db

PawelGutkowski:
So far I have created only 2 dump files for RA-2.3.1 (`openmrs-distro-2.3.1.sql `and `openmrs-distro-2.3.1-failing.sql`). First dump was created before Context initalization, it is not full database of fresh installation (about 100 of 143 tables). It doesn't contain any global properties or security/language settings, and it works perfect. 
Second dump (failing one) is full database dump, and it makes first run fail: 
- initalsetup page shows message : "Invalid user and/or password : admin Unable to complete the startup", 
- then HTTP 500 with error message "Problem accessing openmrs/index.html, reason NullPointerException",
- [part of logs and stacktrace](http://pastebin.com/2LkXxDXC)

Second run is without errors, but demo data is not available. To recreate the behaviour, pull this branch and run `setup` for RA-2.3.1 with parameter `-DdbSql=classpath://openmrs-distro-2.3.1-failing.sql`.
I think that first dump file(partial DB) is fine, it significantly reduces time of setup, and from what I've seen most of global properties are overwritten anyway, but if it is necessary, I can tinker around and try to make it work with full DB dump.


rkorytkowski:
It didn't fail for me... The error is strange... Is it possible that you had some left overs from a previous setup (in db or server directory)? Could you please try again?

rkorytkowski:
Rebased against master in https://github.com/openmrs/openmrs-sdk/pull/24

rkorytkowski:
Sorry you are right. I'm getting the error. dbSQL was not picked up in maven 3.0.5 properly thus I ran it with openmrs-distro-2.3.1.sql first time..

PR #24, opened 2016-07-01
SDK-105

rkorytkowski:
Rebased against master.

https://issues.openmrs.org/browse/SDK-105


maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -218,6 +243,53 @@ private boolean connectMySqlDatabase(Server server) throws MojoExecutionExceptio
...
+                getLog().error(e.getMessage());
+            }
+        }
+    }
+
+    private Reader getSqlScriptReader(String sqlScriptPath){
+        InputStream stream;
+        if(sqlScriptPath.startsWith(CLASSPATH_SCRIPT_PREFIX)){
+            String resource = sqlScriptPath.replace(CLASSPATH_SCRIPT_PREFIX, "");
+            stream = (Setup.class.getClassLoader().getResourceAsStream(resource));
```
rkorytkowski: Will it work if I provide the sql in https://github.com/openmrs/openmrs-distro-referenceapplication/tree/master/package/src/main/resources and specify db.sql=classpath://openmrs-distro-2.4.sql in https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/package/src/main/resources/openmrs-distro.properties ?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -218,6 +243,53 @@ private boolean connectMySqlDatabase(Server server) throws MojoExecutionExceptio
...
+        } else {
+            File scriptFile = null;
+            File relative = new File(System.getProperty("user.dir"), sqlScriptPath);
```
rkorytkowski: I think you can just use new File(sqlScriptPath) for both aboslute and relative. The class will determine if it is relative or absolute based on checking if the path starts from '/' or 'c:/'

rkorytkowski:
Let's just remove the ...failing.sql and settle on the sql before context initialization...

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
PawelGutkowski: It willl not, I didn't consider this use case, I will fix it

rkorytkowski:
Merged in #27

PR #25, opened 2016-07-01
SDK-106 rename installation.properties if exist

AdamGrzybkowski:


maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
```diff
@@ -130,6 +130,12 @@ public static void setServersPath(String serversPath){
     public static boolean hasServerConfig(File dir) {
         if (dir.exists()) {
             File properties = new File(dir, SDKConstants.OPENMRS_SERVER_PROPERTIES);
+            if(!properties.exists()){
```
rkorytkowski: Could you please add a comment that it is for backwards compatibility with 2.x and below

AdamGrzybkowski: @rkorytkowski  done

PR #31, opened 2016-07-05
SDK-101 Build goal, and tests

AdamGrzybkowski:


maven-plugin/src/main/java/org/openmrs/maven/plugins/Build.java
```diff
@@ -0,0 +1,171 @@
...
+        if (!tempFolder.exists()) {
+            tempFolder.mkdir();
+        }
+        Model tempModel = createModel();
+        Set<Project> watchedModules = server.getWatchedProjects();
+        for(Project module: watchedModules){
+            Path newLink = Paths.get(new File(tempFolder, module.getArtifactId()).getAbsolutePath());
+            Path existingfile = Paths.get(module.getPath());
+            try {
+                Files.createSymbolicLink(newLink, existingfile);
```
rkorytkowski: It seems createSymbolicLink requires administrator privileges on Windows. We should catch IOException here and simply copy directory if creating symbolic link fails...

integration-tests/src/test/resources/integration-test/module1/pom.xml
```diff
@@ -0,0 +1,18 @@
...
+  <url>http://maven.apache.org</url>
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>3.8.1</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
```
rkorytkowski: You must create actual modules using openmrs-sdk:create-platform-module / openmrs-sdk:create-project...

maven-plugin/src/main/java/org/openmrs/maven/plugins/Build.java
AdamGrzybkowski: ok, thanks for that :)

rkorytkowski: Maybe setup platform instead of Reference Application to speed it up a bit?

rkorytkowski: Also platform modules should build faster than ref app modules...

AdamGrzybkowski: ok :) I will try it

AdamGrzybkowski: @rkorytkowski building platform modules is much faster :) PR updated

PR #32, opened 2016-07-05
SDK-104 pull rebase if git project present

AdamGrzybkowski:
@rkorytkowski I've found one problem with Jgit which I can't resolve. To reproduce it please try to pull changes from project directory with one commit that will be in conflict with master. (no uncommieted changes). 
after doing rebase --abort there are some modified files. I was trying to stash them but it didn't work

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Pull.java
@@ -1,7 +1,6 @@
 package org.openmrs.maven.plugins;
 
 import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Model;
 import org.apache.maven.plugin.BuildPluginManager;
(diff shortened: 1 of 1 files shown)


rkorytkowski:
http://stackoverflow.com/questions/134882/undoing-a-git-rebase

Basically you need to remember a hash of the commit before doing rebase and then just git reset --hard previous-hash-here. In such a case you also have to remember about unstashing any previously stashed changes.

AdamGrzybkowski:
ok, by mistake i was running pull command from openmrs-core directory, where the current branch should be updated to the lates 1.11.x not master. 
But even that it should work and not leave modified files. Reseting to the userBranch didn't help. 

![image]
(http://i.imgur.com/rMTfaDl.png)

rkorytkowski:
Regarding pulling for openmrs-core and other repos... I oversimplified by always pulling from master...  We want to pull from a branch on upstream, which is named same as a remote tracking branch, but also give an option to specify a branch i.e. openmrs-sdk:pull -Dfrom=1.11.x. If no tracking branch is specified for the current branch then ask the user what is the name of the branch he wants to pull from and set it as a remote tracking branch for next calls with 'git branch -u origin/1.11.x 1.11.x'. To determine the remote tracking branch in git you do 'git status', which says something along the lines: 'Your branch is ahead of 'origin/1.11.x' by 19 commits'. origin/1.11.x is the remote tracking branch set for this branch so in our pull we would replace origin with upstream and use the 1.11.x branch.

rkorytkowski:
I couldn't replicate the rebase --abort issue locally... Let's have a look at it on your machine tomorrow.

PR #45, opened 2016-07-12
SDK-113 Verify JDK version on setup

tmarzeion:


maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -191,10 +191,20 @@ else if(isCreatePlatform) {
...
         }
 
-        wizard.promptForJavaHomeIfMissing(server);
 
         server.setUnspecifiedToDefault();
         configureVersion(server, isCreatePlatform);
+
+        String platformVersion = server.getPlatformVersion();
+        if (platformVersion.charAt(0) == 1) {
+            wizard.showMessage("Note: JDK 1.7 is needed for platform version " + platformVersion + ".");
```
tmarzeion: @rkorytkowski What about earlier versions of JDK? Should v1.* platform work on JDK v1.6?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -305,25 +305,70 @@ public void promptForJavaHomeIfMissing(Server server) {
...
+                    promptForJavaHomeIfMissing(server);
+                }
             }
         }
+    }
 
-
+    private String extractJavaVersionFromPath(String path) {
+        path = path.substring(0, path.length()-4);
+        File jdkReleasePropertiesFile = new File(path, "release");
```
tmarzeion: Thats how JDK version is extracted from JDK directory.
Works for Oracle releases.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
rkorytkowski: Nope, we dropped support for 1.6.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
rkorytkowski: That's not correct. You should execute 'java -version' using processbuilder to get the version.

rkorytkowski: Actually sorry, I should have asked if your solution works for OpenJDK? If yes, then it's correct.

tmarzeion: Fixed :)

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -305,25 +307,74 @@ public void promptForJavaHomeIfMissing(Server server) {
...
+
+        String result = null;
 
+        try {
+            final Process process = processBuilder.start();
+            BufferedReader stdInput = new BufferedReader(new InputStreamReader(process.getErrorStream()));
+
+            result = stdInput.readLine();
+        } catch (IOException e) {
+            e.printStackTrace();
```
rkorytkowski: Do not print stack trace.... Rethrow as RuntimeException.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -305,25 +307,74 @@ public void promptForJavaHomeIfMissing(Server server) {
...
+
+        ProcessBuilder processBuilder = new ProcessBuilder(commands);
+
+        processBuilder.directory(new File(path + File.separator + "bin"));
+
+        String result = null;
 
+        try {
+            final Process process = processBuilder.start();
+            BufferedReader stdInput = new BufferedReader(new InputStreamReader(process.getErrorStream()));
```
rkorytkowski: Why getErrorStream?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -305,25 +307,74 @@ public void promptForJavaHomeIfMissing(Server server) {
...
+
+            result = stdInput.readLine();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
 
+        return result.split("\"")[1];
```
rkorytkowski: Could you please use some regex instead? It seems error prone if the ouput changes even slightly.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
tmarzeion: Fixed by using `processBuilder.redirectErrorStream(true)`

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
tmarzeion: Fixed

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
tmarzeion: Sorry, I've probably missed that.
Fixed

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -317,19 +321,78 @@ public void promptForJavaHomeIfMissing(Server server) {
...
+                    server.setJavaHome(null);
+                }
+                else if (System.getProperty("java.version").startsWith(requiredJdkVersion)) {
+                    server.setJavaHome(null);
+                }
+                else {
+                    showMessage("Your JAVA_HOME version doesn't fit platform requirements:");
+                    showMessage("JAVA_HOME version: " + System.getProperty("java.version"));
+                    showMessage("Required: " + requiredJdkVersion);
+                    promptForJavaHomeIfMissing(server);
```
rkorytkowski: Let's make it less strict and require only java 8 for platform 2.x, but allow using java 8 for platform 1.x with additional confirmation from the user "It is not recommended to run OpenMRS platform {versionHere} on JDK 8. Would you like to select the recommended JDK 7 instead? [Y/n]:"

tmarzeion: Fixed

PR #48, opened 2016-07-14
SDK-109 followup - fix creating refapp module

PawelGutkowski:
reference application module archetype doesn't need additional archetypes, I have overlooked that earlier.


PawelGutkowski:
It's just small fix, CI passed, I will merge it. If anyone has any comments, I will gladly adress them in next PR, but it shouldn't be necessary.

PR #54, opened 2016-07-18
SDK-86 followup: fix deploying module with flags

PawelGutkowski:
Small fix, we noticed that if user wants to deploy module running sdk with flags(`-DartifactId=...` etc.) interactive mode with prompt asking what sdk has to do showes up anyway. This PR fixes it.


PawelGutkowski:
CI passed, I will merge it

PR #63, opened 2016-07-26
SDK-130 Support dockerized MySQL

PawelGutkowski:
Messages may need some adjustments, what do You think @rkorytkowski ?


docker-maven-plugin/src/main/java/org/openmrs/maven/plugins/AbstractDockerMojo.java
```diff
@@ -0,0 +1,96 @@
...
+            String answer = prompter.prompt(message);
+            if(StringUtils.isBlank(answer)) return prompt(message);
+            else return answer;
+        } catch (PrompterException e) {
+            throw new RuntimeException("Failed to prompt", e);
+        }
+    }
+
+    protected void showMessage(String message) {
+        System.out.print(message);
```
rkorytkowski: I'd use println

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
```diff
@@ -22,6 +22,7 @@
     // attributes
     public static final String PROPERTY_SERVER_ID = "server.id";
     public static final String PROPERTY_DB_DRIVER = "connection.driver_class";
+    public static final String PROPERTY_DOCKER_MYSQL = "dockerized.mysql";
```
rkorytkowski: let's just store here docker container id and rename to db.docker.container.id

pom.xml
```diff
@@ -17,6 +17,7 @@
         <module>archetype-module-platform</module>
         <module>maven-plugin</module>
         <module>integration-tests</module>
+        <module>docker-maven-plugin</module>
```
rkorytkowski: Let's move it above maven-plugin to better visualise the order of building.

PR #73, opened 2016-08-06
SDK-134 Allow setting values in openmrs-runtime.properties

AdamGrzybkowski:
@rkorytkowski 
I've changed the property options a little bit. You can specify three values in distro.properties:
1) property.%propertyName%=value
2) property.%propertyName%.prompt=value
3) property.%propertyName%.default=value
With option 1 specified wizard won't prompt for this value, if there will be only option 2 wizard will ask for value with specified prompt(option 3 is optional for default value)

The properties are then saved in openmrs-server.properties as %propertyName%=value.

The property.%propertyName% form distro can be override be passing -DpropertyName=value in command line.

Unfortunately putting properties in openmrs-servers.properties and setting them as system properties doesn't seem to solve @mogoodrich problem, I guess pih.config is not read as system value. I managed to setup PIH only with pih.config in runtime.properties. 

 Lastly I've changed promptForValueIfMissingWithDefault() method in wizzard, so the user does not have to write(in option nr 2, mentioned above) something like "Pleas specify value for someProperty, default: %s". The default String will be added automatically if defValue is specified.

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
@@ -2,7 +2,6 @@ package org.openmrs.maven.plugins.model;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.maven.plugin.MojoExecutionException;
-import org.apache.maven.project.MavenProject;
 
 
 import java.io.File;
@@ -11,7 +10,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -22,6 +21,7 @@ import java.util.Set;
  */
 public class DistroProperties {
 
+    public static final String PROPERTY_PROMPT_KEY = "property.%s.prompt";
     private static final String DEAFAULT_FILE_NAME = "openmrs-distro-%s.properties";
     private static final String TYPE_OMOD = "omod";
     private static final String TYPE_WAR = "war";
@@ -31,7 +31,8 @@ public class DistroProperties {
     private static final String GROUP_ID = "groupId";
     public static final String DISTRO_FILE_NAME = "openmrs-distro.properties";
     private static final String DB_SQL = "db.sql";
-
+    public static final String PROPERTY_DEFAULT_VALUE_KEY = "property.%s.default";
+    public static final String PROPERTY_KEY = "property.%s";
 
 
     private Properties properties;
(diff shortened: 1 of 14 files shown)


mogoodrich:
Thanks @AdamGrzybkowski.  So everything in openmrs-server.properties gets set as system properties on startup?

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
```diff
@@ -191,6 +194,42 @@ public String getName(){
...
+        return propertiesNames;
+    }
+
+    private String extractPropertyName(String key) {
+        int beginIndex = key.indexOf(".");
+        int endIndex = key.indexOf(".prompt");
+        if(endIndex != -1){
+            return key.substring(beginIndex + 1, endIndex);
+        } else  {
+            endIndex = key.indexOf(".default");
```
rkorytkowski: .default should be as last so it should be rather key = StringUtils.stripStart(key, "property."); if (key.endsWith(".default")) { key = StringUtils.stripEnd(key, ".default")  } else { key = StringUtils.stripEnd(key, ".prompt"); }

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
```diff
@@ -712,6 +712,10 @@ public void setContainerId(String containerId) {
         setParam(PROPERTY_DOCKER_MYSQL, containerId);
     }
 
+    public void setPropertyValue(String propertyName, String value){
+        setParam("property."+propertyName+".default", value);
```
rkorytkowski: Shouldn't you be calling setParam("property." + propertyName, value) to be concise with distro.properties?

rkorytkowski:
@mogoodrich, to be precise every property starting from "property." gets set as a system property.

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
AdamGrzybkowski: I think StringUtils.stripStart is not the best solution for this. If you try to do StringUtils.stripStart("property.pih.config", "property.") as a result you will get ih.config. 

Secondly I don't think the order is important here ? We just want all the property names, no matter if it is prompt or default

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
AdamGrzybkowski: It definitely should, my mistake

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
rkorytkowski: My point is that you can have property.pih.default.config.default and your code breaks.
Not sure why stripStart would cut off 'p'... Seems like a bug or it treats '.' in a special way...

AdamGrzybkowski: According to docs [https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#stripStart(java.lang.String,%20java.lang.String)](StringUtils.stripStart) it works as expected. It deletes every char from the given string, not the whole one if present. 

I have changed the code and added your example to distro properties test file.

PR #78, opened 2016-08-09
SDK-140 dbUri default with docker machine ip

AdamGrzybkowski:
I will test it tomorrow on Windows machine


AdamGrzybkowski:
@rkorytkowski We tested it on mac and it works

PR #79, opened 2016-08-10
SDK-138 Fix container searching, set fixed docker api version

PawelGutkowski:
Had to fix Docker API version, because docker-java library is not compatibile with new version 1.24, so it has to be explicitly declared in docker client initialization.


docker-maven-plugin/src/main/java/org/openmrs/maven/plugins/AbstractDockerMojo.java
```diff
@@ -64,7 +65,13 @@ protected void resolveDocker() {
                 dockerHost = prompt("Please specify you Docker Machine host URL (format is: 'tcp://{docker-machine url}')","");
             }
         }
-        docker = DockerClientBuilder.getInstance(dockerHost).build();
+
+        DockerClientConfig config = DockerClientConfig.createDefaultConfigBuilder()
+                .withDockerHost(dockerHost)
+                .withApiVersion("1.23")
```
PawelGutkowski: @rkorytkowski on Ubuntu 14.04 works without problems.

PR #84, opened 2016-08-16
SDK-85 fixed node cmd for windows

AdamGrzybkowski:
After creating OWA, all files are removed except node directory.

I'm coping both npm and npm.cmd scripts to the OWA, but I'm not sure about that. Should I copy npm.cmd on Windows os and npm on  non windows os ?

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/CreateProject.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/CreateProject.java
@@ -3,6 +3,7 @@ package org.openmrs.maven.plugins;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.SystemUtils;
 import org.apache.maven.archetype.ArchetypeManager;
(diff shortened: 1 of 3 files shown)


rkorytkowski:
Copying both is desired.

maven-plugin/src/main/java/org/openmrs/maven/plugins/CreateProject.java
```diff
@@ -447,6 +450,20 @@ private void createOwaProject() throws MojoExecutionException {
...
+    private void addHelperScripts(String path) {
+        File npmCmd = new File(path, "npm.cmd");
+        URL npmCmdSrc = getClass().getClassLoader().getResource("npm.cmd");
+        File npm = new File(path, "npm");
+        URL npmSrc = getClass().getClassLoader().getResource("npm");
+        try {
+            FileUtils.copyURLToFile(npmSrc, npm);
+            FileUtils.copyURLToFile(npmCmdSrc, npmCmd);
+        } catch (IOException e) {
+            e.printStackTrace();
```
rkorytkowski: Never do this. You can change your IDE to use new RuntimeExcetion(e); in the try catch template.

PR #85, opened 2016-08-16
SDK-136 modified deploy module to work with server properties

AdamGrzybkowski:
@rkorytkowski This is not finished yet. Any feedback will be appreciated :)

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
@@ -18,7 +18,7 @@ import java.util.*;
 /**
  * Class for Server model
  */
-public class Server {
+public class Server extends BaseSdkProperties {
     // attributes
     public static final String PROPERTY_SERVER_ID = "server.id";
     public static final String PROPERTY_DB_DRIVER = "connection.driver_class";
@@ -45,8 +45,6 @@ public class Server {
 
     private static String serversPath = System.getProperty("user.home")+File.separator+SDKConstants.OPENMRS_SERVER_PATH;
 
-    private Properties properties;
-
     private File propertiesFile;
 
     private File serverDirectory;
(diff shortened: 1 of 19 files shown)


AdamGrzybkowski:
@rkorytkowski 
I've implemented old functionality using properties from server, but now I have the following questions:
1) Should we keep war.openmrs... properties in server properties ? If yes what about openmrs.platform.version, delete it ? 
2) I've deleted userModules, because we keep current modules versions in server and distro properties. How openmrs-sdk:reset should work now ? 

I still have to add server validation for servers created with sdk version below 3.2.

AdamGrzybkowski:
@rkorytkowski In the last commit I've added support for downgrading modules when deploying distro. I think if we support deleting we should also support downgrading. Openmrs Core can be downgraded to, but not deleted.

AdamGrzybkowski:
@rkorytkowski I've resolved conflicts

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/ModuleProperties.java
```diff
@@ -0,0 +1,206 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+import java.util.Set;
+
+/**
+ *
+ */
+public class ModuleProperties {
```
PawelGutkowski: This name is a bit misleading, maybe we could name it "BaseSdkProperties". I guess this class could be abstract as well

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/Server.java
```diff
@@ -658,11 +660,17 @@ public void setIncludeDemoData(boolean includeDemoData) {
...
 
-    public String getPlatformVersion(){
+    @Deprecated
+    public String getOpenmrsCoreVersion(){
         return getParam(PROPERTY_PLATFORM);
     }
-    public void setPlatformVersion(String platformVersion){
-        setParam(PROPERTY_PLATFORM, platformVersion);
+
+    public void saveOpenmrsCoreVersion(String version) throws MojoExecutionException {
```
PawelGutkowski: Saving changes in distro properties file, but not changing openmrs-server properties file seems a bit incosistent and can possibly introduce bugs

PR #118, opened 2016-12-06
SDK-179 Fetch different versions of openmrs-distro.properties for the…

AdamGrzybkowski:
@rkorytkowski I've modified it so now all version displayed to the user are released versions of distro-platform. When choosing custom you can fetch openrms war file by providing the version only or different platform-distro that was no displayed by typing org.openmrs.distro:platform:{version} or platform:{version}

If that is fine to you I will edit the wiki once this PR is merged


rkorytkowski:
I'd rather when choosing **custom** you can specify version and SDK tries to fetch org.openmrs.distro:platform:{version}, if it is not available then it fetches just war...

AdamGrzybkowski:
@rkorytkowski  done

PR #120, opened 2016-12-07
SDK-167 - Add -DresetDb feature when overriding existing DB - WIP

tmarzeion:


maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -296,6 +331,33 @@ private boolean connectMySqlDatabase(Server server) throws MojoExecutionExceptio
...
+            connector = new DBConnector(uri, server.getDbUser(), server.getDbPassword(), server.getDbName());
+            DatabaseMetaData md = connector.getConnection().getMetaData();
+            ResultSet rs = md.getTables(server.getDbName(), null, null,new String[] {"TABLE"});
+            List<String> tableList = new ArrayList<>();
+            while (rs.next()) {
+                tableList.add(rs.getString(tableNameColumnIndex));
+            }
+            return tableList;
+        } catch (SQLException e) {
+            return null;
```
rkorytkowski: Why would you ever ignore an exception and just return null?

tmarzeion: Sorry, It seems that I just forgot to handle this :)
Fixed

tmarzeion:
@rkorytkowski 
There is problem with this ticket:
-DdbReset function works just fine, I've checked its logic and database is wiped or not when we want to.
The problem is that when we are "reusing" database that were initialized before (by performing openmrs-sdk:run and going to page for the first  time) it tries to **initialize allready initialized** database and server doesn't really like that.
Do You have any idea how to handle this?

rkorytkowski:
Why you mean by "it tires to initialize already initialized database"? Is it entering a setup wizard/failing/something else?

tmarzeion:
@rkorytkowski It enters setup wizard, but I think he shouldn't. Correct me if I'm wrong

tmarzeion:
There are a lot of liquibase validation errors like this one:
http://pastebin.com/WuCMnST7

tmarzeion:
I've did more research today:
(Chronological order)
1. Setup server with new DB - [LOGS](https://issues.openmrs.org/secure/attachment/48535/SETUP_NEWDB)
2. Run server and go to `localhost:8080/openmrs` to initialize DB, after that I've created new patient - [LOGS](https://issues.openmrs.org/secure/attachment/48537/RUN_NEWDB)
3. Remove server folder from `home/openmrs/<servname>`
4. Setup with using existing DB - [LOGS](https://issues.openmrs.org/secure/attachment/48536/SETUP_EXISTINGDB)
5. Run server and go to`localhost:8080/openmrs` - its the moment where many liquibase errors happen. Whole process takes about 10 minutes, then MySQL connection cannot be estabilished because of  `Too many connections` (Don't look at message that says `Please check your openmrs runtime properties file and make sure you have the correct connection.username and connection.password set` because it seems that its default message that occurs whenever MySQL refuses connection) - [LOGS](https://issues.openmrs.org/secure/attachment/48539/RUN_EXISTINGDB_1)
6. Run server again, now server is running fine, without any unusual errors, I've checked for patient and started visit for him - [LOGS](https://issues.openmrs.org/secure/attachment/48538/RUN_EXISTINGDB_2)

tmarzeion:
@rkorytkowski Done :)

rkorytkowski:
Please resolve conflicts and merge...

PR #124, opened 2016-12-29
SDK-185 - Ability to define parent distro in openmrs-distro.properties

tmarzeion:
TODO:
Thanks @AdamGrzybkowski for help on this ticket :)

- [x] Create logic in `getModuleAndWarProperties` that will create correct properties (with valid property names and skipping default values if present)
- [x] Create tests

### DIFF
--- a/integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
+++ b/integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
@@ -330,6 +330,26 @@ public class SetupIntegrationTest extends AbstractSdkIntegrationTest {
         assertThat(server, serverHasDebugPort("1044"));
     }
 
+    @Test
+    public void setup_shouldInstallWithParentDistroSpecifiedInDistroProperties() throws Exception{
+        final String distroPropertiesDirectoryName = "distro-properties";
+        final String distroPropertiesFileName = "openmrs-distro.properties";
+        final String distroDirectoryParam = "distro";
+
+        addTaskParam(distroDirectoryParam, testDirectory.toString() + File.separator + distroPropertiesDirectoryName + File.separator + distroPropertiesFileName);
+        addMockDbSettings();
+
+        String serverId = UUID.randomUUID().toString();
+        addAnswer(serverId);
+        addAnswer("1044");
+
+        executeTask("setup");
+
+        assertFilePresent( serverId + File.separator + "openmrs-2.0.1.war");
+        assertModulesInstalled(serverId, "htmlformentry-3.3.1.omod");
+        assertFileNotPresent(serverId + File.separator + "modules" + File.separator + "htmlformentry-3.3.0.omod");
+    }
+
     private String readValueFromPropertyKey(File propertiesFile, String key) throws Exception {
 
         InputStream in = new FileInputStream(propertiesFile);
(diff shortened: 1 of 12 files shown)


integration-tests/src/test/resources/integration-test/distro-properties/openmrs-distro.properties
```diff
@@ -0,0 +1,8 @@
+#Fri Dec 30 10:54:43 CET 2016
+version=2.5
+name=Reference Application
+distro.referenceapplication.artifactId=referenceapplication-package
```
rkorytkowski: We want SDK to translate referenceapplication to referenceapplication-package in code automatically so that this line is not needed.

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
```diff
@@ -215,6 +215,26 @@ public void setup_shouldInstallServerWithGivenJavaHomeAndAddJavaHomeToSdkPropert
         assertThat(server, serverHasDebugPort("1044"));
     }
 
+    @Test
+    public void setup_shouldInstallWithParentDistroSpecifiedInDistroProperties() throws Exception{
```
rkorytkowski: Could you please add a test for doing it recursively? e.g setup_shouldInstallWithParentAndGrandparentDistro

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
```diff
@@ -131,6 +133,68 @@ private String extractPropertyName(String key) {
...
+        return getPlatformVersion();
+    }
+
+    private List<Artifact> mergeArtifactLists(List<Artifact> childArtifacts, List<Artifact> parentArtifacts) {
+        List<Artifact> artifactList = new ArrayList<>();
+        artifactList.addAll(childArtifacts);
+        for (Artifact parentArtifact : parentArtifacts) {
+            boolean found = false;
+            for (Artifact childArtifact : childArtifacts) {
+                if (childArtifact.getArtifactId().equals(parentArtifact.getArtifactId())) {
```
rkorytkowski: artifactId is not enough... groupId and packaging must match too

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
AdamGrzybkowski: @rkorytkowski I think this can be done once there is at least one release with distro containing parent. We've tested this making one locally.

integration-tests/src/test/resources/integration-test/distro-properties/openmrs-distro.properties
tmarzeion: Fixed
Its exclusive feature for `referenceapplication` since I didn't see any other distro with `-package` suffix

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
tmarzeion: @rkorytkowski We tested this manually by making fake releases locally. Its difficult to do integration tests for this since we don't have any released distro with specified parent in `openmrs-distro.properties`

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
tmarzeion: Fixed

PR #127, opened 2017-01-03
SDK-187 - Added support for Docker for Windows and Mac

tmarzeion:
@rkorytkowski 
I am checking for **Docker for Windows** by executing HTTP Request to `http://127.0.0.1:2375/` and it works fine, but I am trying to do same for **Docker for Mac** and I don't know whats default URL in this case (I've temporary set it to `tcp://127.0.0.1:2375/`, which probably isn't correct).
It would be nice if You test is on Mac and tell me what should I put into `DEFAULT_HOST_DOCKER_FOR_MAC` constant.
You can build this code on Mac to test it, since You will be prompted for Docker host URL (if `tcp://127.0.0.1:2375/` as default Docker for Mac host address is incorrect)

### DIFF
--- a/docker-maven-plugin/src/main/java/org/openmrs/maven/plugins/AbstractDockerMojo.java
+++ b/docker-maven-plugin/src/main/java/org/openmrs/maven/plugins/AbstractDockerMojo.java
@@ -2,15 +2,14 @@ package org.openmrs.maven.plugins;
 
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.model.Container;
+import com.github.dockerjava.core.DefaultDockerClientConfig;
 import com.github.dockerjava.core.DockerClientBuilder;
 import com.github.dockerjava.core.DockerClientConfig;
-import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.SystemUtils;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.codehaus.plexus.components.interactivity.Prompter;
-import org.codehaus.plexus.components.interactivity.PrompterException;
 
 import java.io.IOException;
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


docker-maven-plugin/src/main/java/org/openmrs/maven/plugins/AbstractDockerMojo.java
```diff
@@ -54,26 +56,32 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+                    dockerHost = prompt("Type updated host url (You can save this property later by calling openmrs-sdk:setup with defined -DdockerHost)");
+                    resolveDocker();
+                }
+                else {
+                    throw new RuntimeException(message, e);
+                }
+            }
+            else {
+                message = "Could not run `docker info`. If you are using 'Docker Toolbox', please make sure you run the SDK command from the 'Docker Toolbox' terminal.";
+            }
```
rkorytkowski: Please change this behavior as follows:

1. IF not Linux: 
Could not connect to Docker at dockerHostUrl. If you are using 'Docker Toolbox', please make sure you run the SDK command from the 'Docker Toolbox' terminal. If the docker host URL is not correct, please specify correct URL by passing the -DdockerHost="tcp://..." parameter.

2. IF Linux:
Could not connect to Docker at dockerHostUrl. If the docker host URL is not correct, please specify new URL by passing the -DdockerHost="tcp://..." parameter.

In both cases we should stop further execution.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -711,26 +713,88 @@ private String getDefaultDbUri(Server server, DockerHelper dockerHelper) {
...
+                    showMessage("Docker Host URL is needed in order to connect to your Docker instance.");
+                    host = promptForValueIfMissingWithDefault("Please specify Docker host URL",
+                            host, "dockerHost", "");
+                }
+            }
+            else if (SystemUtils.IS_OS_MAC_OSX) {
+                showMessage("Checking default Docker for Mac host address:");
+                if (isDefaultDockerForMacPresent()) {
+                    host = DockerHelper.DEFAULT_HOST_DOCKER_FOR_MAC;
+                }
```
rkorytkowski: Docker For Mac seems to be using unix:///var/run/docker.sock 

Check 'docker-machine url' first. It returns "Host is not running" in Docker For Mac. Then use unix://var/run/docker.sock as default.

tmarzeion: @rkorytkowski 

Thanks for code review :)
I've refactored code responsible for `-DdockerHost` property.
Now it behaves like that:

**1. IF Running for first time WITHOUT -DdockerHost specified**
- Try to find docker host by checking for default addresses and assign it to `sdk.properties`
- If docker address is invalid throw exception with message You suggested above

**2. IF Running NOT for first time WITHOUT -DdockerHost specified**
- Use docker host specified as property in sdk.properties
-  If `dockerHost` in `sdk.properties` is not found, behave same as in point **1.**

**3. IF Running WITH -DdockerHost specified WITHOUT value**
- Behave same as in point **1.**

**4. IF Running WITH -DdockerHost specified WITH value**
- Use specified docker host address value and assign it to `sdk.properties`
- If docker address is invalid throw exception with message You suggested above

Notice that customization of dockerHost parameter is now able only in batch mode, I've removed this property from interactive mode assuming that we don't want to confuse people who doesn't use custom docker host addresses.

tmarzeion:
@rkorytkowski 

Thanks for code review :)
I've refactored code responsible for `-DdockerHost` property.
Now it behaves like that:

**1. IF Running for first time WITHOUT -DdockerHost specified**
- Try to find docker host by checking for default addresses and assign it to `sdk.properties`
- If docker address is invalid throw exception with message You suggested above

**2. IF Running NOT for first time WITHOUT -DdockerHost specified**
- Use docker host specified as property in sdk.properties
-  If `dockerHost` in `sdk.properties` is not found, behave same as in point **1.**

**3. IF Running WITH -DdockerHost specified WITHOUT value**
- Behave same as in point **1.**

**4. IF Running WITH -DdockerHost specified WITH value**
- Use specified docker host address value and assign it to `sdk.properties`
- If docker address is invalid throw exception with message You suggested above

Notice that customization of dockerHost parameter is now able only in batch mode, I've removed this property from interactive mode assuming that we don't want to confuse people who doesn't use custom docker host addresses.

PR #129, opened 2017-01-04
SDK-191 - Store OpenMRS data in volume

tmarzeion:
By adding `/usr/local/tomcat/.OpenMRS/modules/` and `/usr/local/tomcat/.OpenMRS/owa/` to `web` service's `volumes` its possible to separate `modules` and `owa` directories content from docker volume, so every container got its own `modules` and `owa` content.

### DIFF
--- a/maven-plugin/src/main/resources/build-distro/docker-compose.yml
+++ b/maven-plugin/src/main/resources/build-distro/docker-compose.yml
@@ -27,6 +27,11 @@ services:
       DB_CREATE_TABLES: 'false'  # change to 'true' if using an empty database
       DB_ADD_DEMO_DATA: 'false'  # change to 'true' if using an empty database
       DB_AUTO_UPDATE: 'false'    # change to 'true' if using an empty database
+    volumes:
+      - web-data:/usr/local/tomcat/.OpenMRS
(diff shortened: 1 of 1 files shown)


maven-plugin/src/main/resources/build-distro/docker-compose.yml
```diff
@@ -27,6 +27,11 @@ services:
       DB_CREATE_TABLES: 'false'  # change to 'true' if using an empty database
       DB_ADD_DEMO_DATA: 'false'  # change to 'true' if using an empty database
       DB_AUTO_UPDATE: 'false'    # change to 'true' if using an empty database
+    volumes:
+      - web-data:/usr/local/tomcat/.OpenMRS
+      - /usr/local/tomcat/.OpenMRS/modules/
+      - /usr/local/tomcat/.OpenMRS/owa/
```
tmarzeion: This is where I found solution for that:
http://stackoverflow.com/a/37898591

tmarzeion: docker-compose volume documentation:
https://docs.docker.com/compose/compose-file/#/volumes-volumedriver

PR #136, opened 2017-01-18
SDK-194 OWA feature should be built with SDK

AdamGrzybkowski:


archetype-submodule-owa/src/main/resources/archetype-resources/pom.xml
```diff
@@ -10,43 +10,16 @@
     <build>
         <plugins>
             <plugin>
-                <groupId>com.github.eirslett</groupId>
-                <artifactId>frontend-maven-plugin</artifactId>
-                <version>1.0</version>
+                <groupId>org.openmrs.maven.plugins</groupId>
+                <artifactId>openmrs-sdk-maven-plugin</artifactId>
+                <version>3.6.2</version>
```
rkorytkowski: Should this be set to a version with that feature implemented? Currently 3.7.1-SNAPSHOT?

AdamGrzybkowski: Should this be SNAPSHOT ? or just 3.7.1 ? I don't know if this version will be changed automatically when releasing as in other poms.

PR #137, opened 2017-01-20
SDK-195 - Build owa uses system version

tmarzeion:
### TODO:
- [x] **Handle root permissions** - Fixed by using `NVM` (https://github.com/creationix/nvm), which installs `npm` and `nodejs` that doesn't need root access.
- [x] **Test new functionalities**

Now there are 4 places where npm and nodejs version could be described (order by priority):
1. Batch argument (described as `npmVersion` and `nodejsVersion`)
2. System version
3. OWA Project version (described as properties: `npm-version` and `nodejs-version` in `package.json`)
4. Default version.

PR #138, opened 2017-01-24
SDK-196 - Added watch-owa goal

tmarzeion:
This commit contains followup to https://github.com/openmrs/openmrs-sdk/pull/137


maven-plugin/src/main/java/org/openmrs/maven/plugins/model/PackageJson.java
```diff
@@ -0,0 +1,111 @@
+package org.openmrs.maven.plugins.model;
```
rkorytkowski: Missing license....

maven-plugin/src/main/java/org/openmrs/maven/plugins/model/NodeDistVersion.java
```diff
@@ -0,0 +1,23 @@
+package org.openmrs.maven.plugins.model;
```
rkorytkowski: Missing license...

maven-plugin/src/main/java/org/openmrs/maven/plugins/WatchOwa.java
```diff
@@ -0,0 +1,57 @@
+package org.openmrs.maven.plugins;
```
rkorytkowski: Missing license...

PR #139, opened 2017-02-20
SDK-205 - Deploy is suggesting actually installed distro versions

tmarzeion:


maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SDKConstants.java
```diff
@@ -115,7 +115,12 @@ public static Artifact getReferenceModule(String version) {
         Artifact artifact = new Artifact("referenceapplication-package", version, Artifact.GROUP_DISTRO, Artifact.TYPE_ZIP);
         artifact.setClassifier("distro");
         return artifact;
+    }
 
+    public static Artifact getDistroModule(String groupId, String artifactId, String version) {
+        Artifact artifact = new Artifact(artifactId, version, groupId, Artifact.TYPE_ZIP);
+        artifact.setClassifier("distro");
```
rkorytkowski: Why setting the distro classifier?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SDKConstants.java
```diff
@@ -115,7 +115,12 @@ public static Artifact getReferenceModule(String version) {
         Artifact artifact = new Artifact("referenceapplication-package", version, Artifact.GROUP_DISTRO, Artifact.TYPE_ZIP);
         artifact.setClassifier("distro");
         return artifact;
+    }
 
+    public static Artifact getDistroModule(String groupId, String artifactId, String version) {
+        Artifact artifact = new Artifact(artifactId, version, groupId, Artifact.TYPE_ZIP);
```
rkorytkowski: Shouldn't this be jar type?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java
```diff
@@ -149,7 +149,13 @@ private void runInteractiveMode(Server server, ServerUpgrader upgrader) throws M
                         server.getName(),
                         server.getVersion()));
 
-                distro = wizard.promptForRefAppVersion(versionsHelper);
+                if (server.getName().equals("Platform") || server.getDistroGroupId() == null || server.getDistroArtifactId() == null) {
+                    // If its impossible to define distro, prompt refapp distro versions
+                    distro = wizard.promptForRefAppVersion(versionsHelper);
+                } else {
+                    // If its possible to define distro, prompt that distro's versions
+                    distro = wizard.promptForDistroVersion(server.getDistroGroupId(), server.getDistroArtifactId().replace("-omod",""), server.getVersion(), server.getName(), versionsHelper);
```
rkorytkowski: Why are you removing "-omod"? Shouldn't we be using artifactId as defined?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SDKConstants.java
tmarzeion: I've removed this one

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SDKConstants.java
tmarzeion: Yup, its JAR by default, fixed

maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java
tmarzeion: There is a little problem since I've used org.openmrs.module:pihmalawi test case for this issue.
Problem is that because of this `distro`'s `groupId` is `org.openmrs.module` (instead of `org.openmrs.distro`) SDK saves its `artifactId` in `server.properties` as `pihmalawi-omod`.
Question is:
Do we want to support that "distro-module" artifacts by stripping `-omod`?

rkorytkowski:
Did you see travis failure?

tmarzeion:
@rkorytkowski Yeah, its caused by problems with servers where integration tests are running

PR #141, opened 2017-02-20
SDK-197 SDK is unnecessarily querying maven repos

AdamGrzybkowski:


maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/VersionsHelper.java
```diff
@@ -80,11 +81,15 @@ private void sortDescending(List<ArtifactVersion> versions){
...
      */
     public String inferVersion(Artifact artifact){
-        List<ArtifactVersion> artifactVersions = getVersions(artifact);
-        if(artifactVersions.size() == 0){
+        if (new Version(artifact.getVersion()).isSnapshot()) {
+            List<ArtifactVersion> artifactVersions = getVersions(artifact);
+            if(artifactVersions.size() == 0){
+                return artifact.getVersion();
+            }
+            return inferVersion(artifact.getVersion(), artifactVersions);
```
rkorytkowski: Actually, why is it calling inferVersion, if the version is set already?

AdamGrzybkowski: Do we allow to no set version in distro properties ?

rkorytkowski: No

AdamGrzybkowski: So should I remove this version inferring or leave it as it is now?

rkorytkowski: @AdamGrzybkowski I don't have enough knowledge about all the callers of this method to answer your question. I said what is the expected behavior of this particular feature so go ahead and decide what changes are needed.

AdamGrzybkowski: @rkorytkowski artifact.getVersion is never null(or it shouldn't be :) ). If it was there would be an exception in this [line](https://github.com/openmrs/openmrs-sdk/blob/master/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/VersionsHelper.java#L58). The only thing this infer method help us is if there was some beta/alpha release and user specified only version ( 2.3 will be changed e.g to 2.3-alpha if 2.3 was not released).  Is there a possibility that there will be some alpha-SNAPSHOT version ? If no then this method is useless now. Let me know and I will apply changes

AdamGrzybkowski:
@rkorytkowski I did delete inferring versions, but left this method. It might be useful in future.

AdamGrzybkowski:
@rkorytkowski PR updated

PR #152, opened 2019-03-15
SDK-230: Allow empty string passwords in JDBC

bgalusha-unl:
https://issues.openmrs.org/browse/SDK-230

During SDK setup, if a user attempts to use an empty password to connect to a database,  JDBC throws a connection error.

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
@@ -884,9 +884,9 @@ public class DefaultWizard implements Wizard {
         server.setDbUser(user);
         //set password
         String dbPassword = promptForPasswordIfMissingWithDefault(
-                "Please specify database password (-D%s)",
-                server.getDbPassword(), "dbPassword", " ");
-        server.setDbPassword(dbPassword);
+                "Note: whitespace-only passwords will be left blank and trailing/leading whitespace will be removed.\nPlease specify database password (-D%s)",
+                server.getDbPassword(), "dbPassword", "");
+        server.setDbPassword(dbPassword.trim());
     }
 
     /**


bgalusha-unl:
Some clarification on the tests we performed:

We are able to successfully connect to MySQL with a database user that has no password. This is done by simply hitting enter when prompted for a database password in the setup wizard. Setting up the SDK with a database user that has a password is still possible and has not been affected so far as we know.

However, it is now impossible to use a password consisting only of whitespace. A single space (which was the default before) is no longer accepted. This is because the StringUtils.isBlank method treats whitespace-only strings as "blank." We have deemed this behavior acceptable, but it may be worth documenting in https://wiki.openmrs.org/display/docs/OpenMRS+SDK#OpenMRSSDK-Basicusage

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -885,7 +885,7 @@ public void promptForDbCredentialsIfMissing(Server server) {
         //set password
         String dbPassword = promptForPasswordIfMissingWithDefault(
                 "Please specify database password (-D%s)",
-                server.getDbPassword(), "dbPassword", " ");
+                server.getDbPassword(), "dbPassword", "");
```
RishavR: Solution is acceptable. However since it doesn't allow white space only passwords (which users are unlikely do anyway), we must explicitly mention in OpenMRS SDK documentation.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -885,7 +885,7 @@ public void promptForDbCredentialsIfMissing(Server server) {
...
         server.setDbPassword(dbPassword);
```
RishavR: I would suggest doing: 

server.setDbPassword(dbPassword.trim()); 

in order to ensure that the password does not have any leading or trailing blank spaces which might cause a authentication issue later on.

Reyano132:
Hey @bgalusha-unl , can you please squash your commits.
Hint : https://wiki.openmrs.org/display/docs/Pull+Request+Tips

bgalusha-unl:
Update: We added a short message to the password prompt warning about whitespace passwords.

dkayiwa:
Can you include a link to the ticket as advised at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

bgalusha-unl:
I think I added a link to the ticket.

Also, is it necessary to include tests for this PR?

gracepotma:
@dkayiwa can this be merged or closed?

PR #157, opened 2019-11-26
SDK-245: Upgrade Tomcat version in SDK

ibacher:
See [SDK-245](https://issues.openmrs.org/browse/SDK-245).

The SDK is using a version of Tomcat that doesn't appropriately handle Jars built with the Java Module System. This PR simply bumps the version of Tomcat to a version in the 7.x series that supports these Jars.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <tomcat.version>7.0.67</tomcat.version>
+        <tomcat.version>7.0.96</tomcat.version>
     </properties>
 
     <repositories>


dkayiwa:
Have you looked at the travis build failure?

ibacher:
@dkayiwa I was hoping that build failure wouldn't show up. It's a regression for this package that's caused by Maven 3.6.2. I.e. the same code succeeds on Maven 3.6.1 and earlier, but fails on 3.6.2 or 3.6.3. The last successful build, [625](https://travis-ci.org/openmrs/openmrs-sdk/builds/613591775#L132), used Maven 3.6.0; the Travis image must have been updated since then, as [build 626](https://travis-ci.org/openmrs/openmrs-sdk/builds/617014015#L129) uses Maven 3.6.2 and thus has this failure.

ibacher:
In any case, the failure is unrelated to this code change and should probably be addressed as a separate issue.

PR #160, opened 2020-02-01
SDK-248 Bundle latest java8 and tomcat7 in Docker

rmkanda:
Refer - https://issues.openmrs.org/browse/SDK-248

Bundle latest java8 and tomcat7 in Docker since the base image from tomcat bundles older version of tomcat and java (also the base image contains outdated and vulnerable packages).


dkayiwa:
Did you get a chance to look at this? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

cintiadr:
Thanks for the pull request, but I'm having trouble understanding the reasoning behind it. 

For maintainability, we prefer to use the [Docker official tomcat image](https://hub.docker.com/_/tomcat/) always. It simplifies a lot our dockerfile and we don't have to reinvent the wheel. 

What's the reason to not do that?

rmkanda:
@cintiadr The official image from tomcat is out of date and contains vulnerable packages. Refer the screenshot below (I can share the detailed scan report if you need).

Since OpenMRS runs on tomcat 7 (nearing EOL) and Java 8 (already EOL). The official images are not actively maintained. So we can't relay on the official images for updates. 

The only option left for us is to package the required packages with latest base image.
BTW, I wrote this Dockerfile by referring the official specs of libraries.  

image

cc: @dkayiwa

rmkanda:
@dkayiwa I have gone through the Pull+Request+Tips doc. Please let know if I am missing any.

dkayiwa:
@rmkanda did you look at the part where it talks about the commit message?

cintiadr:
I still don't think this is a good trade off. 

This will make upgrading and maintaining images (and even for users to modify it) a lot more painful. And what we actually should do is to go to a maintained version of tomcat/java, otherwise we are just being lucky of actually having a exploitable security problem (not the false 'everything is wrong' that docker scans yells). 

Even if we do this change now, the refapp release is months away. Platform has just being released. 

@dkayiwa  what's the plan to go to a maintained version of tomcat and java? Because that's the actually dangerous ones.

cintiadr:
I mean, not sure if using the slim images might make those numbers go down (it might), but it's not my major worry here.

dkayiwa:
I believe we can run on some higher versions of tomcat.
As for Java, we need some reasonable dev efforts.

cintiadr:
Hi @isears , you might want to check this one and give your opinion :)

cintiadr:
>  As for Java, we need some reasonable dev efforts.

i get it, so that's why we need to start as soon as we can. Do we have a plan on doing so?

dkayiwa:
It was attempted here: https://talk.openmrs.org/t/upgrading-from-jdk-8-to-openjdk-11/23181
But i think we should assign the task to a more skilled dev.
@ibacher you usually get attracted to such tickets?

ibacher:
Upgrading the JDK version support is on my radar, but I'm not going to have enough time to devote to it for probably a couple of months. I have explored it a bit [and my thoughts on what needs to happen are here](https://issues.openmrs.org/browse/TRUNK-5496?focusedCommentId=260272&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-260272).

Incidentally, while I can't speak to what vulnerabilities are being reported, the Tomcat 7 images [are still being maintained](https://github.com/docker-library/repo-info/blob/master/repos/tomcat/local/7-jdk8-adoptopenjdk-hotspot.md). Also, while its not ideal to be on 7 (instead of, say, 9), I wouldn't worry too much about upgrading it until an actual EOL is announced.

isears:
@cintiadr thanks for the ping.

This has come up in the past and I remember scanning  both the openmrs container and the tomcat7 container with trivy and finding a 1:1 relationship between openmrs vulnerabilities and tomcat7 vulnerabilities. So we're basically relying on apache to manage that.

> This will make upgrading and maintaining images (and even for users to modify it) a lot more painful. And what we actually should do is to go to a maintained version of tomcat/java, otherwise we are just being lucky of actually having a exploitable security problem (not the false 'everything is wrong' that docker scans yells).

Huge +1, basically all I came here to say. @rmkanda I'm glad you're looking into this, but the scan results alone are not a good reason to make this change. If you're concerned about these images you would have a lot more impact contacting apache about them.

rmkanda:
@dkayiwa @cintiadr The official images are released from tomcat and openjdk as well. So closing this PR.

rmkanda:
@dkayiwa @cintiadr Please note that the tomcat comes from latest image is 7.99 but the required version mentioned in pom.xm is 7.96. We need to update pom.xml.

dkayiwa:
@rmkanda can you update it?

rmkanda:
Sure @dkayiwa

PR #166, opened 2020-04-21
SDK-252: Updating settings.xml

sherrif10:
https://issues.openmrs.org/browse/SDK-252

 cc @ibacher ,@dkayiwa

### DIFF
--- a/maven-plugin/src/main/resources/settings.xml
+++ b/maven-plugin/src/main/resources/settings.xml
@@ -24,7 +24,7 @@
             </properties>
             <repositories>
                 <repository>
-                    <id>openmrs repo</id>
+                    <id>openmrs-repo</id>
                     <name>OpenMRS Nexus Repository</name>
                     <url>http://mavenrepo.openmrs.org/nexus/content/repositories/public</url>
                 </repository>
--- a/maven-plugin/src/main/resources/settings.xml
+++ b/maven-plugin/src/main/resources/settings.xml
@@ -24,10 +24,22 @@
             </properties>
             <repositories>
                 <repository>
-                    <id>openmrs-repo</id>
+                    <id>openmrs repo</id>
                     <name>OpenMRS Nexus Repository</name>
                     <url>http://mavenrepo.openmrs.org/nexus/content/repositories/public</url>
                 </repository>
+              <repository>
+                <id>archetype</id>
+                <url>http://mavenrepo.openmrs.org/nexus/content/repositories/public</url>
+               <releases>
+                 <enabled>true</enabled>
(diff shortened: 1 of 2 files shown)


maven-plugin/src/main/resources/settings.xml
```diff
@@ -24,9 +24,17 @@
...
                     <name>OpenMRS Nexus Repository</name>
                     <url>http://mavenrepo.openmrs.org/nexus/content/repositories/public</url>
+                     <releases>
+				     <enabled>true</enabled>
+					 <checksumPolicy>fail</checksumPolicy>
+					 </releases>
+					 <snapshots>
+					 <enabled>true</enabled>
+					 <checksumPolicy>warn</checksumPolicy>
+					 </snapshots>
```
ibacher, line 37: Can you please ensure to consistently only use spaces for the formatting here? Also, we shouldn't remove the `openmrs-repo` identifier nor add the policies to it. We should instead create a new repository entry, even though it's referring to the same repository. This allows us to have a specific policy that applies to downloading archetypes, but not necessarily to other Maven artefacts.

pom.xml
```diff
@@ -163,7 +163,7 @@
             <dependency>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-archetype-plugin</artifactId>
-                <version>2.4</version>
+                <version>3.0.0</version>
```
ibacher, line 166: On reflection, I wouldn't change this version right now. There is some brittle code elsewhere in the SDK, which is why the Travis build is failing.

pom.xml
```diff
@@ -186,6 +186,11 @@
                 <artifactId>httpclient</artifactId>
                 <version>4.5.2</version>
             </dependency>
+           
+            <dependency>
+			    <groupId>org.dom4j</groupId>
+			    <artifactId>dom4j</artifactId>
+           </dependency>
```
ibacher, line 193: This is fine to keep in, but please consistently use on spaces for indentation here.

sherrif10: i think it looks ok now

maven-plugin/src/main/resources/settings.xml
```diff
@@ -24,10 +24,22 @@
             </properties>
             <repositories>
                 <repository>
-                    <id>openmrs-repo</id>
+                    <id>openmrs repo</id>
```
ibacher, line 27: This should still be `openmrs-repo`

sherrif10: Sorry @ibacher   for late reply my computer had got issues but am now settled thanks for this

PR #168, opened 2020-04-28
SDK-253 - Update Dockerfile-jre7 and Dockerfile-jre7-bundled to use t…

mseaton:
…he tomcat:7-jre7 base image, rather than the tomcat7 base image, in order to prevent moving from Java 7 to Java 8 for OpenMRS 1.x environments


integration-tests/src/test/resources/integration-test/buildIT/pom.xml
```diff
@@ -9,4 +9,13 @@
...
+
+    <repositories>
+        <repository>
+            <id>openmrs-repo</id>
+            <name>OpenMRS repository</name>
+            <url>http://mavenrepo.openmrs.org/nexus/content/repositories/public</url>
+        </repository>
+    </repositories>
+
 </project>
```
mseaton, line 21: Note: I had to add the repositories section into 2 of the poms in the project to get the integration tests to run.  This is likely due to my having removed my settings.xml file from my .m2 directory recently.  This allows the unit tests to pass without this environmental dependency.

PR #169, opened 2020-04-30
SDK-222: Allow user to specify port to use when default taken

jwnasambu:
Description of what I did:
I worked on an option where by when a port is taken while running openmrs SDK, the SDK should not only suggests un alternative port to be used but also  allow a user to specify a port they want to use. To achieve this I took the input to the wizard prompt and turn it into the port value which is returned by the value of the promptForValueIfMIssingWithDefault() function.

Ticket worked on:
https://issues.openmrs.org/projects/SDK/issues/SDK-222?filter=allopenissues


jwnasambu:
@ibacher , @dkayiwa , @mozzy11 Kindly review my PR at your convenient time am open to any suggestion.

PR #171, opened 2020-05-18
TRUNK-4830 Avoid running legacy liquibase changesets

WolfSchlegel:
This pull request adds two records to the `liquibasechangelog` table.

Without those values updating the database with the new Liquibase snapshots introduced by 2.4.0-SNAPSHOT is failing.

Creating, starting and restarting servers was successfully tested for both OpenMRS 2.3.0 and 2.4.0-SNAPSHOT.


dkayiwa:
Thanks @WolfSchlegel for fixing this. I have tested the sdk and confirmed that all is well. 😊

PR #177, opened 2021-04-07
SDK-273 - Adds a step during the BuildDistro and Setup tasks that att…

mseaton:
…empts to download an owa from Maven if it fails to download from Bintray

@dkayiwa @ibacher and @mogoodrich - pinging you each on this, as I think we're going to be dealing with a number of Bintray related implications.

This isn't intending to be an end-all-be-all solution for fixing Bintray or OWA handling in the SDK, but this covers the two main use cases I am most concerned about in the PIH workflow, which is:

* Setting up a new SDK instance of our distribution, if it contains owas in the openmrs-distro.properties file, and having these OWAs installed into the owa directory as indicated (pulling from either Bintray or Maven).  This is the "Setup" task.

* Building a distribution zip for deployment to test/production servers, which uses the SDK maven plugin to pull down the war file, modules, and owas defined in openmrs-distro.properties so that they can be packaged up.  This is the "BuildDistro" task.

We likely need more extensive changes throughout the SDK.  But I don't want that job to delay these improvements, if possible.

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -14,6 +14,7 @@ import org.openmrs.maven.plugins.model.DistroProperties;
 import org.openmrs.maven.plugins.model.Server;
 import org.openmrs.maven.plugins.model.Version;
 import org.openmrs.maven.plugins.utility.DistroHelper;
+import org.openmrs.maven.plugins.utility.OwaHelper;
 import org.openmrs.maven.plugins.utility.Project;
 import org.openmrs.maven.plugins.utility.SDKConstants;
 
@@ -262,31 +263,11 @@ public class BuildDistro extends AbstractTask {
         List<Artifact> owas = distroProperties.getOwaArtifacts(distroHelper, targetDirectory);
         if (!owas.isEmpty()) {
             wizard.showMessage("Downloading OWAs...\n");
-            for (Artifact owa: owas) {
-                installOwa(owasDir, owa);
-            }
-        }
-    }
-
-    private void installOwa(File owasDir, Artifact owa) throws MojoExecutionException {
-        // Support existing Bintray artifacts
-        try {
+            OwaHelper owaHelper = new OwaHelper();
             OpenmrsBintray openmrsBintray = new OpenmrsBintray(getProxyFromSettings());
-            openmrsBintray.downloadOWA(owasDir, owa.getArtifactId(), owa.getVersion());
-        }
-        catch (Exception e) {
-            // If file is not found in bintray, and exception will be thrown.  Try to find it in Maven repo.
-            moduleInstaller.installModule(owa, owasDir.getAbsolutePath());
-            File downloadedFile = new File(owasDir, owa.getArtifactId() + "-" + owa.getVersion() + "." + owa.getType());
-            if (!downloadedFile.exists()) {
-                throw new MojoExecutionException("Unable to download OWA from Bintray or Maven", e);
-            }
-            File renamedFile = new File(owasDir, owa.getArtifactId() + OpenmrsBintray.OWA_PACKAGE_EXTENSION);
-            try {
-                FileUtils.moveFile(downloadedFile, renamedFile);
-            }
-            catch (IOException ioe) {
-                throw new MojoExecutionException("Unable to move OWA file to " + renamedFile, e);
+            for (Artifact owa: owas) {
+                wizard.showMessage("Downloading OWA: " + owa);
+                owaHelper.downloadOwa(owasDir, owa, openmrsBintray, moduleInstaller);
             }
         }
     }
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -17,7 +17,12 @@ import org.openmrs.maven.plugins.utility.DistroHelper;
 import org.openmrs.maven.plugins.utility.Project;
 import org.openmrs.maven.plugins.utility.SDKConstants;
 
-import java.io.*;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileWriter;
+import java.io.FilenameFilter;
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.util.List;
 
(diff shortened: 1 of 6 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -335,12 +340,34 @@ private void installOWAs(Server server, DistroProperties distroProperties) throw
...
+
+    private void installOwa(File owasDir, Artifact owa) throws MojoExecutionException {
+        // Support existing Bintray artifacts
+        try {
+            OpenmrsBintray openmrsBintray = new OpenmrsBintray(getProxyFromSettings());
+            openmrsBintray.downloadOWA(owasDir, owa.getArtifactId(), owa.getVersion());
+        }
+        catch (Exception e) {
+            // If file is not found in bintray, and exception will be thrown.  Try to find it in Maven repo.
+            moduleInstaller.installModule(owa, owasDir.getAbsolutePath());
```
mogoodrich, line 360: Tangential, but I guess we will want to change the name of this from "moduleInstaller" to "artifactInstaller" at some point? (but agree, let's not do it now)

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -335,12 +340,34 @@ private void installOWAs(Server server, DistroProperties distroProperties) throw
...
             wizard.showMessage("Downloading OWAs...\n");
             for (Artifact owa: owas) {
-                openmrsBintray.downloadOWA(owasDir, owa.getArtifactId(), owa.getVersion());
+                wizard.showMessage("Downloading OWA: " + owa);
+                installOwa(owasDir, owa);
+            }
+        }
+    }
+
+    private void installOwa(File owasDir, Artifact owa) throws MojoExecutionException {
```
mogoodrich, line 352: is there reason we have to duplicate these method twice, in BuildDistro and Setup? Or is there a difference I"m not noticing?

mogoodrich (approved):
Had a couple questions, but generally LGTM @mseaton ... though I'm not very familiar with the SDK so it would be good to hear from @dkayiwa  or @ibacher  as well...

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -255,12 +260,33 @@ private String buildDistro(File targetDirectory, Artifact distroArtifact, Distro
 
     private void downloadOWAs(File targetDirectory, DistroProperties distroProperties, File owasDir) throws MojoExecutionException {
         List<Artifact> owas = distroProperties.getOwaArtifacts(distroHelper, targetDirectory);
-        OpenmrsBintray openmrsBintray = new OpenmrsBintray(getProxyFromSettings());
-
         if (!owas.isEmpty()) {
             wizard.showMessage("Downloading OWAs...\n");
             for (Artifact owa: owas) {
-                openmrsBintray.downloadOWA(owasDir, owa.getArtifactId(), owa.getVersion());
+                installOwa(owasDir, owa);
```
dkayiwa, line 266: Did you intentionally change the name from `downloadOWA` to `installOwa`?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -335,12 +340,34 @@ private void installOWAs(Server server, DistroProperties distroProperties) throw
...
+            File downloadedFile = new File(owasDir, owa.getArtifactId() + "-" + owa.getVersion() + "." + owa.getType());
+            if (!downloadedFile.exists()) {
+                throw new MojoExecutionException("Unable to download OWA from Bintray or Maven", e);
+            }
+            File renamedFile = new File(owasDir, owa.getArtifactId() + OpenmrsBintray.OWA_PACKAGE_EXTENSION);
+            try {
+                FileUtils.moveFile(downloadedFile, renamedFile);
+            }
+            catch (IOException ioe) {
+                throw new MojoExecutionException("Unable to move OWA file to " + renamedFile, e);
```
dkayiwa, line 370: Did you intend to pass `ioe` instead of `e` in `new MojoExecutionException`?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
mseaton: Yes, agreed, but didn't want to overcomplicate this PR.  I think the entire SDK could do with a "3.0" version that makes significant updates without worry for backwards compatibility.

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
mseaton: I didn't want to duplicate it, but didn't really know the right place to put it.  I could add it to "AbstractTask", but that didn't seem to match the model that was followed elsewhere.  I thought about putting it in OwaHelper, but it would have required more wiring around of things.  The same was true for other Helpers and ModuleInstaller.  Ultimately, it was easiest to just duplicate the code, knowing that we'll likely change down the road.  I'll see if I can change this, as I agree it is not ideal to have the same logic duplicated.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
mseaton: yes?  it's a private method, so don't see the harm.  do you think downloadOWA is a better method name?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
mseaton: Yes, good catch

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
mseaton: OK, I pushed an update that moves the new logic into a new method within OwaHelper

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
mseaton: @dkayiwa renamed back to downloadOWA in new push

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
mseaton: Fixed in latest commit.

mseaton:
@dkayiwa, @mogoodrich, @ibacher - going to merge this in as I pushed some fixes aimed at addressing your concerns.  Can address any subsequent comments and requests as follow-up commits.

PR #185, opened 2021-05-28
SDK-281: SDK should install Microfrontends specified in openmrs-distro.properties

brandones:
https://issues.openmrs.org/browse/SDK-281


maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.twdata.maven.mojoexecutor.MojoExecutor.executionEnvironment;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.goal;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.groupId;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.plugin;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.version;
+
+public class NodeHelper {
+    private static final String FRONTEND_MAVEN_GROUP_ID = "com.github.eirslett";
+    private static final String FRONTEND_MAVEN_ARTIFACT_ID = "frontend-maven-plugin";
+    private static final String FRONTEND_MAVEN_VERSION = "1.12.0";
```
ibacher, line 29: Probably best to stick with 1.11.3, if possible. 1.12.0 requires Maven 3.6.0+, which we actually don't have on the CI build agents (IIRC, we tried to upgrade to Maven 3.6 and it broke some builds).

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Set;
+
+public class SpaInstaller {
+
+    static final String BAD_SPA_PROPERTIES_MESSAGE = "Distro properties file contains invalid 'spa.' elements. " +
+            "Please check the distro properties file and the specification. " +
+            "Parent properties cannot have their own values (i.e., if 'spa.foo.bar' exists, 'spa.foo' cannot be assigned a value). " +
+            "Duplicate properties are not allowed.";
+    static final String NODE_VERSION = "16.2.0";
+    static final String NPM_VERSION = "7.6.2";
```
ibacher, line 27: Node 16.2.0 was bundled with NPM 7.13.0. Is there a reason to keep pinned to the older version?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
```diff
@@ -0,0 +1,122 @@
...
+        try {
+            out = new FileWriter(file);
+            jsonObject.writeJSONString(out);
+        }
+        catch (IOException e) {
+            throw new MojoExecutionException(e.getMessage());
+        }
+        finally {
+            IOUtils.closeQuietly(out);
+        }
```
ibacher, line 119:
```suggestion
        try (FileWriter out = new FileWriter(file)) {
        	jsonObject.writeJSONString(out);
        }
        catch (IOException e) {
        	throw new MojoExecutionException(e.getMessage(), e);
        }
```

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
```diff
@@ -0,0 +1,122 @@
...
+        } else {
+            if (!jsonObject.containsKey(keys[0])) {
+                jsonObject.put(keys[0], new JSONObject());
+            }
+            Object childObject = jsonObject.get(keys[0]);
+            if (!(childObject instanceof JSONObject)) {
+                throw new RuntimeException(BAD_SPA_PROPERTIES_MESSAGE +
+                        " Also please post to OpenMRS Talk and include this full message. If you are seeing this, there has been a programming error.");
+            }
+            String childKeys = StringUtils.join(Arrays.copyOfRange(keys, 1, keys.length), ".");
```
ibacher, line 103:
```suggestion
            String childKeys = propertyKey.substring(propertyKey.indexOf(".") + 1)
```

ibacher (commented):
Some tests would be good. We should probably also document somewhere that this doesn't support MFs with `.` in their name, though that's admittedly unlikely.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
brandones: Any reason for this? The StringUtils version will be easier to refactor, since it leaves the only reference to `propertyKey` in the function in the first line. They seem about equally simple to me.

ibacher: Mostly, just for less data copying, but for the expected one-value nest, I don't think there's much of a difference really.

maven-plugin/pom.xml
```diff
@@ -228,6 +228,13 @@
             <artifactId>exec-maven-plugin</artifactId>
             <version>1.5.0</version>
         </dependency>
+
+        <!-- SPA Microfrontends -->
+        <dependency>
+            <groupId>com.googlecode.json-simple</groupId>
+            <artifactId>json-simple</artifactId>
+            <version>1.1.1</version>
+        </dependency>
```
ibacher, line 237: Is there a reason for favouring this over the Gson library already included?

brandones: I didn't realize there was already a JSON library! Fixing.

mseaton:
@brandones - all in all this looks good to me on a quick review.  My main observation might be around naming - do we want all of this to be referred to as "spa" (spa. properties and Spa* classes).  Is "spa" still the preferred term?  I think of this more as "mfe" than "spa" these days.  Interested in your and others take on this.

ibacher:
I'd almost prefer `frontend` or `ui` (but these would be weird if we ever decide to replace the MFE stuff); I like `spa` better than `mfe`, though, as I think of "microfrontend" as an architectural pattern, which is a weird thing to inflict as a naming convention on end-users.

brandones:
Yeah we already have `openmrs-module-spa`, which is the other thing a beginning implementer is likely to be aware of. I agree with @ibacher both on the benefit and risk of `frontend`. I think that's the other good option.

brandones:
What's the verdict guys? I don't think there's a clear-cut right answer here. Input from @jonathandick , who has put some thought into "branding," would also be welcome.

mogoodrich:
I think naming conventions are important, but I struggle with this one... I like the idea of a more generic name like "ui" or "frontend" (but though that could break if SPA becomes used for more than the front end).

Having a naming convention around a technical approach, (ie single-page-app) seems bad in some cases, but on the other hand, if 10 years down the road we come up with a different UI using a different technology, they both could happily run in parallel if they aren't both named something like "frontend".

Don't know if that helps or adds to the debate?  :)  (I'm fine with sticking with Spa if others think it is the right choice).

mseaton:
A generic term like "frontend" or "ui" is fine _iff_ there is enough information available in the artifact for the SDK or another deployment tool to know what to do with it, independent of technology.  However, there are a lot of "frontend" and "ui" technologies out there - including jsp, gsp, etc.  What exists today is we have "war.", "omod.", "owa."... and these all imply that these are artifacts of a specific type and deployment location into an OpenMRS server.  What is the nature of the artifacts that these goals are aiming to support, and can we make some naming conventions that makes that explicit?

brandones:
@mseaton Per our conversation yesterday, wanna hit the button?

mseaton (approved):
This looks good to me.  Others?

ibacher (approved):

PR #186, opened 2021-05-28
Upgrade Java to 1.8

brandones:


brandones:
Do people use the SDK in production? I thought it was just a development tool, and was considering bumping this straight to Java... something recent

ibacher:
@brandones See [this talk thread](https://talk.openmrs.org/t/getting-error-unsupported-major-minor-version-52-0-when-running-server-with-openmrs-sdk-plugin-3-13-6/29251/4) for why I'm hesitant on this.

brandones:
Oh! And in that thread you link to this commit, which happened in the past, and then was reverted... https://github.com/openmrs/openmrs-sdk/commit/479c3e570484d3619b897e171dc82defc9bb08aa .

Can the source version be bumped safely, while leaving the target version alone?

ibacher:
Unfortunately not. The target can be more recent than the source, but not the other way around.

PR #187, opened 2021-05-28
Remove interfaces

brandones:
As mentioned in Talk: https://talk.openmrs.org/t/interfaces-and-implementations-in-openmrs-sdk/33597

AFAICT these interfaces don't add anything, they're just extra weight that makes it slightly harder to debug and refactor.


mseaton (approved):
Looks fine to me, no issues on my end with making these changes.

brandones:
This seems to have broken the SDK pretty substantially. It appears that the Prompter is now not getting wired(?) into the Wizard. Any command that prompts for input generates

```
Caused by: java.lang.NullPointerException
    at org.openmrs.maven.plugins.utility.Wizard.prompt (Wizard.java:234)
```

which line is

```java
            return prompter.prompt("\n" + textToShow);
```

which indicates that `prompter` is null. Prompter is declared as a property of Wizard with

```java
    @Requirement
    Prompter prompter;
```

Reverting would be the Right Thing To Do™, but there will be some merge difficulties from 726160663ae5e403b034f4fcc1c154ddee90c28e . I'll see if I can figure out how to fix this quickly.

It seems like Prompter and `@Requirement` is part of [Plexus](https://codehaus-plexus.github.io/), which seems to be a legacy IoC system? I have no idea how that is supposed to work.

Suggestions/ideas/help welcome.

brandones:
In the POM, I see that the plugin

```xml
                <groupId>org.codehaus.plexus</groupId>
                <artifactId>plexus-component-metadata</artifactId>
```

is being used with no `version` specifier. This is interesting, because while versions [up to 2.1.0](https://mvnrepository.com/artifact/org.codehaus.plexus/plexus-component-metadata) are available in Maven Central, it appears that the OpenMRS Maven repo (which I think is that my settings are configured to use) only has up to 1.0-beta-3.0.5.

ibacher:
@brandones Arrrgh... I should've tested this locally before merging. Yeah, Plexus is a weird thing that never really saw any adoption outside of Maven, but which is also fundamental to how Maven works. I guess that's why we had these useless interfaces. I think I've reverted things successfully, or, at least it works for me. I haven't actually tried out the SPA-specific stuff though.

On the `plexus-component-metadata` thing, it's probably helpful to know that Maven POMs can have a hierarchy (through the `<parent/>` element and the [parent POM for this](https://github.com/openmrs/openmrs-sdk/blob/92344a5f984b1679c08fb044d30f75e685b2bbb7/pom.xml#L365) limits it to exactly 1.0-beta-3.0.5.

Updating these various bits is definitely on the TODO list.

brandones:
Got it, thanks for taking care of that @ibacher . I've just verified that the SPA stuff still works as expected.

PR #192, opened 2021-06-22
SDK-288  SDK build-distro should build SPA application

brandones:
https://issues.openmrs.org/browse/SDK-288

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -75,6 +75,9 @@ public class BuildDistro extends AbstractTask {
 
     /**
      * @parameter  property="bundled" default-value="false"
+     *
+     * Instead of creating a `modules` folder in the distro directory, will put modules inside
+     * the war file in `/webapp/src/main/webapp/WEB-INF/bundledModules`
      */
     private boolean bundled;
 
@@ -200,6 +203,7 @@ public class BuildDistro extends AbstractTask {
 
         String distroName = adjustImageName(distroProperties.getName());
         File web = new File(targetDirectory, WEB);
+        web.mkdirs();
 
         moduleInstaller.installModules(distroProperties.getWarArtifacts(distroHelper, targetDirectory), web.getAbsolutePath());
         renameWebApp(web);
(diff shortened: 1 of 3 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -208,11 +212,12 @@ private String buildDistro(File targetDirectory, Artifact distroArtifact, Distro
             try {
                 ZipFile warfile = new ZipFile(new File(web, OPENMRS_WAR));
                 File tempDir = new File(web, "WEB-INF");
+                tempDir.mkdir();
                 moduleInstaller.installModules(distroProperties.getModuleArtifacts(distroHelper, targetDirectory),
                         new File(tempDir, WAR_FILE_MODULES_DIRECTORY_NAME).getAbsolutePath());
 
                 File owasDir = new File(tempDir, "bundledOwas");
-                owasDir.mkdirs();
```
dkayiwa, line 215: Was the change from `mkdirs` to `mkdir` intentional?

brandones: Yeah, when I did the refactor of the `setup` task I eventually realized that a `owasDir.mkdirs()` call was actually accidentally being used to create the server directory itself. This made it very hard to reason about what was happening when. That gave me the belief that `mkdirs` should be used sparingly. It opens the door for code to get messy and dependent on what should be side-effects.

dkayiwa: How about `web.mkdirs()` above?

brandones: Since that's the "top" of the directory tree that `BuildDistro` is concerned with, I have less faith that its parent directories have been created already.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -75,6 +75,9 @@
 
     /**
      * @parameter  property="bundled" default-value="false"
+     *
+     * Instead of creating a `modules` folder in the distro directory, will put modules inside
```
dkayiwa, line 79: What would be the advantage of not creating a modules folder in the distro directory?

brandones: I have no idea. It was implemented in https://issues.openmrs.org/browse/SDK-177 . My comment is just there to explain what it does.

PR #193, opened 2021-07-14
SPA installer should handle array-type config parameters correctly

brandones:
This adds support for SPA parameters that take array arguments. Right now this is just `configUrls`.

### DIFF
--- a/integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
+++ b/integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
@@ -164,6 +164,7 @@ public class SetupIntegrationTest extends AbstractSdkIntegrationTest {
         assertFilePresent(serverId + File.separator + "frontend" + File.separator + "openmrs-esm-patient-chart-app-3.0.0");
         String indexContents = FileUtils.readFileToString(new File(testDirectory.getAbsolutePath(), indexFilePath));
         assertThat(indexContents, containsString("apiUrl: \"notopenmrs\""));
+        assertThat(indexContents, containsString("configUrls: [\"foo\"]"));
 
         Server.setServersPath(testDirectory.getAbsolutePath());
         Server server = Server.loadServer(serverId);
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
@@ -86,7 +86,7 @@ public class SpaInstaller {
     /**
      * Add a line from the properties file to the new JSON object. Creates nested objects as needed.
      * Known array-valued keys are parsed as comma-delimited arrays; e.g.,
-     *   `configUrl=qux` becomes `{ "configUrl": ["qux"] }` because `configUrl` is known to be array-valued
+     *   `configUrls=qux` becomes `{ "configUrls": ["qux"] }` because `configUrls` is known to be array-valued
      *
      * @param jsonObject the object being constructed
      * @param propertyKey
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
@@ -32,12 +32,10 @@ public class SpaInstaller {
 
     private DistroHelper distroHelper;
 
-    public SpaInstaller(MavenProject mavenProject,
-                        MavenSession mavenSession,
-                        BuildPluginManager pluginManager,
-                        DistroHelper distroHelper) {
+    public SpaInstaller(DistroHelper distroHelper,
+                        NodeHelper nodeHelper) {
         this.distroHelper = distroHelper;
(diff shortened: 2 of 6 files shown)


integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
```diff
@@ -164,6 +164,7 @@ public void setup_shouldInstallServerFromDistroPropertiesDir() throws Exception{
         assertFilePresent(serverId + File.separator + "frontend" + File.separator + "openmrs-esm-patient-chart-app-3.0.0");
         String indexContents = FileUtils.readFileToString(new File(testDirectory.getAbsolutePath(), indexFilePath));
         assertThat(indexContents, containsString("apiUrl: \"notopenmrs\""));
+        assertThat(indexContents, containsString("configUrls: [\"foo\"]"));
```
mseaton, line 167: It seems odd to me that we wouldn't test this with more than one array parameter, in order to test things like parsing the comma-separated syntax, etc.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
```diff
@@ -83,14 +83,33 @@ private JsonObject convertPropertiesToJSON(Map<String, String> properties) throw
         return result;
     }
 
+    /**
+     * Add a line from the properties file to the new JSON object. Creates nested objects as needed.
+     * If the value contains commas, it is assumed to represent an array of strings. Known array-valued
+     * parameters are always parsed as arrays of strings. e.g.
+     *   `foo=bar,baz` becomes `{ "foo": ["bar", "baz"] }`
+     *   `configUrl=qux` becomes `{ "configUrl": ["qux"] }` because `configUrl` is known to be array-valued
```
mseaton, line 91: configUrl -> configUrls

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
```diff
@@ -83,14 +83,33 @@ private JsonObject convertPropertiesToJSON(Map<String, String> properties) throw
...
+     */
     private void addPropertyToJSONObject(JsonObject jsonObject, String propertyKey, String value) {
         String[] keys = propertyKey.split("\\.");
         if (keys.length == 1) {
             if (jsonObject.has(keys[0])) {
                 throw new RuntimeException(BAD_SPA_PROPERTIES_MESSAGE +
                         " Encountered this error processing a property containing the key '" + keys[0] + "' and with value " + value);
             }
-            jsonObject.addProperty(keys[0], value);
+            if (value.contains(",") || "configUrls".equals(keys[0])) {
```
mseaton, line 104: Are you sure we want to make anything an array if it contains a comma?  Are there use cases where we want this, or possible use cases where we don't?

mseaton (commented):
A few thoughts / suggestions.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/SpaInstaller.java
brandones: I have no idea, it's just speculative future-proofing. It seems more likely to me that we might add other array-valued parameters than that we will ever add a parameter which can include a `,`. But I could be wrong about that, and am open to the idea of removing this functionality.

mseaton: Yes, that is true, but in that case they array will also need to support single-value arrays like "configUrls", and so will also need to be hard-coded as you had to do with "configUrls", so having the generic behavior doesn't really add value or make sense.

brandones: Good point. Removing it.

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java
brandones: Yeah at present there's only one distro.properties file in the tests that includes the `spa.` properties, and the single array value seems more like an edge case which someone might overlook in implementing an array-values parser. Based on my experience having initially implemented this in a way that didn't support single array values :)

Could certainly add SPA building to another distro file but it would slow the tests down substantially.

ibacher: Maybe we could test the functionality of this in a unit test rather than adding a new integration test?

brandones: That is extremely sensible

brandones: Added some unit tests for `SpaInstaller`. Got halfway to adding unit test support for `Setup` too, until I realized that's not what we want here. But maybe that would be useful at some point.

mseaton (approved):

ibacher (approved):
Thanks @brandones! LGTM.

PR #195, opened 2021-07-23
Refactoring things

ibacher:
This is probably too big of a PR to get an effective review of, but this does a bunch of refactoring to make the SDK code base hopefully a little more maintainable.

Highlights:

* Removed duplicate code where possible
* Centralised dependency versions so all plugins are using the same versions
* Updated the versions of most plugins; all are not findable on Maven central
* Integrated the Wizard with JLine, which allows for a better experience when typing long strings
* Adds support for Postgres to a couple of commands that weren't properly supported before

Although the commit dates are recent, that's just due to rebasing. I've actually been using most of these changes for the past few weeks.


ibacher:
Re: build failures: this seems very similar to the issues we've been having with Bamboo, i.e., the part that's failing no longer exists in this branch.

dkayiwa:
@ibacher just in case you did not see the above. 😊

ibacher:
@dkayiwa Sorry... I forgot to mention that that should've been fixed in b8e4276. I got a little too aggressive when trying to fix the NPE...

PR #200, opened 2021-11-22
SDK-293: Refactor SDK so Tomcat is launched in a separate process

ibacher:
Ok, so the problem this is trying to solve is that the SDK launches an embedded version of Tomcat, which means that Tomcat inherits the SDK classpath, including (and this is the real problem) the MySQL driver. The newer version of the MySQL driver (8+) returns metadata that the very old version of Liquibase in pre-2.4 does not understand.

It might've been possible to (somehow) provide Tomcat with a relatively "clean" classpath, but instead I've opted to just run Tomcat in a fork (the code to do so was already in the SDK) and from a separate Maven project, which ensures that the classpath pollution is minimal. Additionally, spinning out the Tomcat part as a separate Maven plugin allows us to continue to support Tomcat 7 (which is necessary for versions of OpenMRS that use the Legacy UI but don't include the fix for TRUNK-6040 / LUI-169).


ibacher:
This isn't perfect, but I'm trying to rapidly patch the issues with the 4.2.0 release of the SDK. In particular, there's a fair bit of code duplication between the two `run-tomcat` commands (possibly unavoidable) and the OpenMRSSDKRedirector should probably be genericised rather than copied twice, but because it's installed in the root context, I wanted to limit what I was passing.

Finally the construction for determining what version of Tomcat to use should be simplified. Right now, to ensure you are using Tomcat 9, you need to use a command like:

```
mvn openmrs-sdk -DrunGoal=org.openmrs.maven.plugins:openmrs-sdk-tomcat9-maven-plugin:4.3.0-SNAPSHOT:run-tomcat
```

ibacher:
This is, unfortunately, another one of those PRs that is probably too big to fully review, but...

mseaton:
I gave this a look over and it generally looks great to me @ibacher  - thanks for taking this on!

PR #204, opened 2022-01-25
SDK-296 Setup produces "${project.basedir}" directory when run with SPA modules in distro.properties

brandones:
https://issues.openmrs.org/browse/SDK-296

The erroneous directory was being created because `setup` is not generally run in the context of a Maven project, and the maven-frontend-plugin was trying to use the Maven project build directory. This changes it to install node and the npm cache in a temporary directory.

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
@@ -6,10 +6,10 @@ import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.project.MavenProject;
 import org.twdata.maven.mojoexecutor.MojoExecutor;
 
-import java.io.File;
-import java.nio.file.Paths;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 
 import static org.twdata.maven.mojoexecutor.MojoExecutor.artifactId;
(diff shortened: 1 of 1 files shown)


ibacher:
This is probably fine. Let me confirm that this doesn't break things in the CI environment (the whole reason why that hack is there is to get things working on CI).

mseaton (approved):
LGTM, assuming @ibacher comes back with an indication that CI is fine.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -29,27 +29,27 @@
...
 	public NodeHelper(MavenProject mavenProject,
 			MavenSession mavenSession,
 			BuildPluginManager pluginManager) {
 		this.mavenProject = mavenProject;
 		this.mavenSession = mavenSession;
 		this.pluginManager = pluginManager;
+		try {
+			this.tempDir = Files.createTempDirectory("openmrs-sdk-node");
+		} catch (IOException e) {
+			throw new RuntimeException(e);
```
ibacher, line 44:
```suggestion
			throw new IllegalStateException("Exception while trying to create temporary directory for node", e);
```

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -29,27 +29,27 @@
...
+	private Path tempDir;
 	
 	public NodeHelper(MavenProject mavenProject,
 			MavenSession mavenSession,
 			BuildPluginManager pluginManager) {
 		this.mavenProject = mavenProject;
 		this.mavenSession = mavenSession;
 		this.pluginManager = pluginManager;
+		try {
+			this.tempDir = Files.createTempDirectory("openmrs-sdk-node");
```
ibacher, line 42: Node, NPM, and the cache all get cleaned-up but not the tempDir itself.

```suggestion
			this.tempDir = Files.createTempDirectory("openmrs-sdk-node");
			this.tempDir.toFile().deleteOnExit();
```

brandones:
Thanks @ibacher !

PR #210, opened 2023-02-01
SDK-300 - Frontend builds should allow one to configure node and npm …

mseaton:
…version


ibacher (approved):
Other than I think we're stuck on 1.11.3 of the frontend plugin to keep compatibly with JDK 7, LGTM. Sorry for the delayed review.

PR #216, opened 2023-04-03
Develop

RishabhDixit1:


maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -46,8 +46,6 @@ public class Setup extends AbstractServerTask {
 
 	public static final String SETUP_SERVERS_PROMPT = "You can setup the following servers";
 
-	private static final String CLASSPATH_SCRIPT_PREFIX = "classpath://";
```
prateek27das, line 49: CLASSPATH_SCRIPT_PREFIX provides a convenient and consistent way to reference resources within a project's classpath. I think it shouldn't be removed. I hope you will look into it.

brandones:
Please don't create meaningless PRs. You are making a mess that other people have to clean up.

PR #219, opened 2023-04-03
Sdk 304: Optimising setModuleProperties method 

prateek27das:
Instead of calling the equals() method on the groupId and type fields of the newModule object, I am calling it on the constant values Artifact.GROUP_MODULE and TYPE_JAR. This is because constants are more likely to be interned and thus comparison with them may be faster than comparison with variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,22 @@ For more details visit [the project page](https://wiki.openmrs.org/display/docs/
 
 ## Requirements
  * Maven 3.x
+    
+    The Latest Version of Apache Maven may Collide with the JDK 8. Try using Apache Maven 3.6.3
+
+
+ * JDK 8
+
  
 ## Installation
+* Maven 3.x 
+   
+Mac:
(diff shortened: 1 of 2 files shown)


ibacher (changes requested):
Please review the PR title; It would be better to be a little more descriptive.

I don't love adding the Mac-specific instructions to the README. Maybe there's an appropriate change page on the Wiki? It's not a bad idea to list JDK 8. AFAIK, we support the latest version of Maven building and working with the SDK.

finally, it's best if each PR does one thing, so updates to the README disconnected from the changes to the code means this should probably be two PRs.

prateek27das:
Thanks for your review. I got the idea and I will try to change the pr according to that. 

So I was trying to build the project in my local system (Mac). But I was getting test failures. Then I figured out that Maven 3.9.1(New Version) was colliding with the Java 8. Then I changed the Maven version to 3.6.3. After the changes I managed to built the sdk in my local System without any test failures.

brandones:
This does nothing and is stale

PR #224, opened 2023-06-22
SDK-306: Integrated Micro Frontend Tooling Setup in OpenMRS SDK - Reactoring

wikumChamith:
This is to address issues with https://github.com/openmrs/openmrs-sdk/commit/a41bab68919d12acedbf721a02247be09eadb72d

Jira ticket: https://issues.openmrs.org/browse/SDK-306


dkayiwa:
@wikumChamith can you look at the build failures?

wikumChamith:
> @wikumChamith can you look at the build failures?

@dkayiwa I think they got fixed with this. Is there a way to re-run checks without pushing again??

wikumChamith:
> > @wikumChamith can you look at the build failures?
> 
> @dkayiwa I think they got fixed with [this](https://github.com/openmrs/openmrs-sdk/pull/225). Is there a way to re-run checks without pushing again??

@dkayiwa I merged master into this.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -779,7 +779,7 @@ private Map<String, String> getDistroVersionsOptionsMap(Set<String> versions, Ve
 	private Map<String, String> getO3VersionsOptionsMap(VersionsHelper versionsHelper,
 			String optionTemplate, String artifactTemplate) {
 		Map<String, String> optionsMap = new LinkedHashMap<>();
-		Artifact artifact = new Artifact("referenceapplication-distro", "3.0.0-SNAPSHOT", "org.openmrs.distro", "zip");
+		Artifact artifact = new Artifact("referenceapplication-distro", null, "org.openmrs.distro", "zip");
```
dkayiwa, line 782: I remember suggesting passing NULL for the version parameter and you said it was not allowed.

wikumChamith: Apologies for the confusion. When evaluating with the debugger I encountered an exception when using null. So I decided not to use null. Upon further investigation, I realized that even though an exception gets thrown, the code was still able to fetch the versions list.

dkayiwa: Can we avoid the exception being thrown?

wikumChamith: [@dkayiwa](https://github.com/dkayiwa) I am sorry, I have made a mistake. I accidentally ran
 mvn org.openmrs.maven.plugins:openmrs-sdk-maven-plugin:4.6.0-SNAPSHOT:setup instead of mvn org.openmrs.maven.plugins:openmrs-sdk-maven-plugin:4.7.0-SNAPSHOT:setup when testing new code changes.
We can't still use null. If we pass null we get the following error.
[ERROR] Failed to execute goal org.openmrs.maven.plugins:openmrs-sdk-maven-plugin:4.7.0-SNAPSHOT:setup (default-cli) on project openmrs-sdk: Failed to setup server: For artifact {org.openmrs.distro:referenceapplication-distro:null:}: The version cannot be empty.
I am changing the parameter to "version".

wikumChamith: @dkayiwa I am really sorry, I have made a mistake. I accidentally ran 
`mvn org.openmrs.maven.plugins:openmrs-sdk-maven-plugin:4.6.0-SNAPSHOT:setup` instead of `mvn org.openmrs.maven.plugins:openmrs-sdk-maven-plugin:4.7.0-SNAPSHOT:setup` when testing new code changes.

We can't still use `null`. If we pass null we get the following error.

`[ERROR] Failed to execute goal org.openmrs.maven.plugins:openmrs-sdk-maven-plugin:4.7.0-SNAPSHOT:setup (default-cli) on project openmrs-sdk: Failed to setup server: For artifact {org.openmrs.distro:referenceapplication-distro:null:}: The version cannot be empty.`

I am changing the parameter to `"version"`.

dkayiwa:
@wikumChamith can we also let someone choose the version of the openmrs assemble and build tools to use? The default can remain `latest` as it currently is. But allow someone to override it.
I guess it will help solve problems like this: https://talk.openmrs.org/t/cannot-build-the-o3-reference-ui/39959 where he is using the default latest which is unstable because of the current significant frontend rewrite and related changes. So it would be great to have an option of using stable previous versions. You can create a separate pull request for this.

`npm exec -- openmrs@latest --version`

wikumChamith:
> @wikumChamith can we also let someone choose the version of the openmrs assemble and build tools to use? The default can remain `latest` as it currently is. But allow someone to override it. I guess it will help solve problems like this: https://talk.openmrs.org/t/cannot-build-the-o3-reference-ui/39959 where he is using the default latest which is unstable because of the current significant frontend rewrite and related changes. So it would be great to have an option of using stable previous versions. You can create a separate pull request for this.
> 
> `npm exec -- openmrs@latest --version`

@dkayiwa should these get included in the distro.properties?

dkayiwa:
> should these get included in the distro.properties?

Can i also pass it in as a command line argument?

wikumChamith:
> > should these get included in the distro.properties?
> 
> Can i also pass it in as a command line argument?

Then we should add new maven parameters for them. @dkayiwa could you create a new ticket for this with more details?? We can discuss this there.

dkayiwa:
Yes you can create a ticket.

PR #232, opened 2023-07-02
SDK-315: Enhance the documentation system for all plugins

wikumChamith:
This updates the output from openmrs-sdk:help.

* Update the documentation with the current capabilities of plugins.

* Add documentation to the plugins that are currently not included in the documentation.

Jira ticket: https://issues.openmrs.org/browse/SDK-315


maven-plugin/src/main/resources/help.yaml
```diff
@@ -24,9 +24,21 @@ help:
         desc: "Add demo data when setting up a platform server."
       - name: "Dfile"
         desc: "Path to installation.properties file for batch setup."
+      - name: "DdockerHost"
+        desc: "Docker host address"
+      - name: "DdbReset"
+        desc: "If 'true' reset the database"
+      - name: "DjavaHome"
+        desc: "Path to JDK Version"
```
dkayiwa, line 32: Path to JDK version or home?

maven-plugin/src/main/resources/help.yaml
```diff
@@ -24,9 +24,21 @@ help:
...
+      - name: "Ddebug"
+        desc: "Port number for enabling remote debugging (defaults to '1044')."
```
dkayiwa, line 34: Is this really a port number or a flag to determine whether we are using debug mode?

maven-plugin/src/main/resources/help.yaml
```diff
@@ -36,6 +48,10 @@ help:
         desc: "Artifact id of an artifact, which you want to deploy."
       - name: "Dversion"
         desc: "Version of an artifact, which you want to deploy."
+      - name: "Ddistro"
+        desc: "OpenMRS Distribution to set up in the format 'groupId:artifactId:version'. You can skip groupId, if it is 'org.openmrs.distro'. You can also give Path to distro.properties file"
+      - name: "Dowa"
+        desc: "Owa property in the format 'owa-name:version'"
```
dkayiwa, line 54: Which property? Version or something else?

maven-plugin/src/main/resources/help.yaml
wikumChamith: It is the debug port. https://github.com/openmrs/openmrs-sdk/blob/12d3ba529b963ab1e8b81e55ce195e883d8facb1/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java#L433-L448

maven-plugin/src/main/resources/help.yaml
wikumChamith: I found the template for this parameter from here: https://github.com/openmrs/openmrs-sdk/blob/12d3ba529b963ab1e8b81e55ce195e883d8facb1/maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java#L113-L119

maven-plugin/src/main/resources/help.yaml
wikumChamith: I used this: https://github.com/openmrs/openmrs-sdk/blob/12d3ba529b963ab1e8b81e55ce195e883d8facb1/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java#L109-L114

I think it should be JAVA_HOME so I changed it in the new commit.

PR #237, opened 2023-07-11
Rename spaCoreVersion with appShellVersion

wikumChamith:


dkayiwa:
Did you test and confirm that we can now pass the parameter as appShellVersion and its respected?

wikumChamith:
Yes. It is working.

PR #243, opened 2023-07-27
Changing the default value of ignorePeerDependencies to true

wikumChamith:


ibacher (approved):

ibacher:
Could we also update the integration test to either default this to false or use a newer version or `openmrs`?

ibacher:
Thanks for the quick work!

PR #244, opened 2023-07-28
SDK-319: Improve the build-distro command by allowing per-version cutomizations.

wikumChamith:
This generates a Docker image based on Java 8 for OpenMRS versions 2.0-2.3 and Java 11 base image for 2.4 and newer versions.

JIRA: https://issues.openmrs.org/browse/SDK-319


maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -307,7 +307,7 @@ private void copyDockerfile(File targetDirectory, DistroProperties distroPropert
 				copyBuildDistroResource("Dockerfile-jre7", new File(targetDirectory, "Dockerfile"));
 			}
 		} else {
-			if (isPlatform2point5AndAbove(platformVersion)) {
+			if(isBetweenPlatform2point0and2point3(platformVersion)) {
```
ibacher, line 310: Have we tested this to verify that it works?

maven-plugin/src/main/resources/build-distro/web/Dockerfile-tomcat8-jre11
```diff
@@ -0,0 +1,23 @@
+FROM tomcat:8-jre11
```
ibacher, line 1: It would probably be best if we could replace these Dockerfiles with something based on openmrs-core, probably not too dissimilar from what we do in the distro-referenceapplication repo: https://github.com/openmrs/openmrs-distro-referenceapplication/blob/main/Dockerfile (I'd actually like to swap out the Dockerfile there for a version built by the SDK).

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
wikumChamith: Yes, I manually tested this.

maven-plugin/src/main/resources/build-distro/web/Dockerfile-tomcat8-jre11
wikumChamith: Are we also replacing the existing docker files? https://github.com/openmrs/openmrs-sdk/tree/master/maven-plugin/src/main/resources/build-distro/web

wikumChamith: @ibacher, @dkayiwa any insights on how we should do this??

ibacher: I'm just imaging having a Dockerfile like:

```docker
FROM openmrs/openmrs-core:nightly-amazoncorretto-11

COPY openmrs.war /openmrs/distribution/openmrs_core
COPY modules /openmrs/distribution/openmrs_modules
COPY owa /openmrs/distribution/openmrs_owa
```

We'd also have to likely update the script [here](https://github.com/openmrs/openmrs-core/blob/master/startup-init.sh) to handle the O3 SPa.

PR #245, opened 2023-08-15
Adding an error if the server path contains a space.

wikumChamith:
Setting up servers is problematic when the server file path contains spaces. So this adds an error message for server directories with spaces.

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
@@ -681,6 +681,10 @@ public class Setup extends AbstractServerTask {
 			throw new MojoExecutionException(
 					"Cannot create server: directory with name " + serverDir.getName() + " already exists");
 		}
+		else if (serverDir.getAbsolutePath().contains(" ")) {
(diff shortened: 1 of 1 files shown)


dkayiwa:
Should it be only the server? Or anything in the path that has spaces?

wikumChamith:
> Should it be only the server? Or anything in the path that has spaces?

@dkayiwa I think this going to only be problematic with the server path.

dkayiwa:
During the conference hackathon, one of the participants had a user account with space hence leading to a path that has space. The server id was fine.

wikumChamith:
> During the conference hackathon, one of the participants had a user account with space hence leading to a path that has space. The server id was fine.

@dkayiwa This code checks the whole server path. So this covers both the user name and server name.

eg: serverDir.getAbsolutePath() = /home/wikum/openmrs/server new

wikumChamith:
@dkayiwa do you think we should have a separate error message for invalid usernames??

dkayiwa:
All we need is communicate to the user about the cause of the failure being any space along the full path from the beginning to the end.

wikumChamith:
> All we need is communicate to the user about the cause of the failure being any space along the full path from the beginning to the end.

@dkayiwa doesn't the current error message convey that??

PR #255, opened 2023-09-20
SDK-321: Optimizing SDK Cache Management

wikumChamith:

- Add the enableDataSaving parameter to the setup-sdk plugin. This parameter enables the data-saving mode. When using this mode, the SDK will reuse the existing cache folder rather than creating a new one.

- In normal mode, SDK will remove the cache folder after running the setup.

## Issue I worked on

see https://issues.openmrs.org/browse/SDK-321

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/SetupSDK.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/SetupSDK.java
@@ -2,6 +2,7 @@ package org.openmrs.maven.plugins;
 
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
 import org.openmrs.maven.plugins.model.SdkStatistics;
 import org.openmrs.maven.plugins.model.Server;
 import org.openmrs.maven.plugins.utility.SDKConstants;
@@ -13,9 +14,16 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.Path;
+import java.util.Properties;
+
+import static org.openmrs.maven.plugins.utility.PropertiesUtils.getSdkProperties;
+import static org.openmrs.maven.plugins.utility.PropertiesUtils.savePropertiesChangesToFile;
 
 @Mojo(name = "setup-sdk", requiresProject = false)
 public class SetupSDK extends AbstractTask {
+
+	@Parameter(defaultValue = "false", property = "enableDataSaving")
+	private boolean enableDataSaving;
 
 	private static final String SUCCESS_TEMPLATE = "SDK installed successfully, settings file: %s";
 
@@ -46,6 +54,10 @@ public class SetupSDK extends AbstractTask {
 		catch (IOException e) {
 			throw new MojoExecutionException(e.getMessage());
 		}
+		Properties sdkProperties = getSdkProperties();
+		File sdkPropertiesFile = Server.getServersPath().resolve(SDKConstants.OPENMRS_SDK_PROPERTIES).toFile();
+		sdkProperties.put("enableDataSaving", String.valueOf(enableDataSaving));
+		savePropertiesChangesToFile(sdkProperties, sdkPropertiesFile);
 	}
 
 	private void initSdkStatsFile() throws MojoExecutionException {
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/SetupSDK.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/SetupSDK.java
@@ -22,8 +22,13 @@ import static org.openmrs.maven.plugins.utility.PropertiesUtils.savePropertiesCh
 @Mojo(name = "setup-sdk", requiresProject = false)
 public class SetupSDK extends AbstractTask {
 
-	@Parameter(defaultValue = "false", property = "enableDataSaving")
-	private boolean enableDataSaving;
+	/**
+	 * Controls whether the SDK reuse the Node cache after the setup.
+	 * By default, this is set to 'false' to save storage space.
+	 * If set to 'true', the SDK will reuse the existing NPM cache
+	 */
+	@Parameter(defaultValue = "false", property = "reuseNodeCache")
+	private boolean reuseNodeCache;
 
 	private static final String SUCCESS_TEMPLATE = "SDK installed successfully, settings file: %s";
 
@@ -56,7 +61,7 @@ public class SetupSDK extends AbstractTask {
 		}
 		Properties sdkProperties = getSdkProperties();
 		File sdkPropertiesFile = Server.getServersPath().resolve(SDKConstants.OPENMRS_SDK_PROPERTIES).toFile();
-		sdkProperties.put("enableDataSaving", String.valueOf(enableDataSaving));
+		sdkProperties.put("reuseNodeCache", String.valueOf(reuseNodeCache));
 		savePropertiesChangesToFile(sdkProperties, sdkPropertiesFile);
 	}
 
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
@@ -5,6 +5,7 @@ import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugin.BuildPluginManager;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.project.MavenProject;
+import org.openmrs.maven.plugins.model.Server;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.twdata.maven.mojoexecutor.MojoExecutor;
@@ -15,9 +16,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Properties;
 
-import static org.openmrs.maven.plugins.utility.PropertiesUtils.getSdkProperties;
 import static org.twdata.maven.mojoexecutor.MojoExecutor.artifactId;
 import static org.twdata.maven.mojoexecutor.MojoExecutor.configuration;
 import static org.twdata.maven.mojoexecutor.MojoExecutor.element;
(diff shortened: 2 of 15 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/SetupSDK.java
```diff
@@ -13,10 +14,17 @@
...
+import java.util.Properties;
+
+import static org.openmrs.maven.plugins.utility.PropertiesUtils.getSdkProperties;
+import static org.openmrs.maven.plugins.utility.PropertiesUtils.savePropertiesChangesToFile;
 
 @Mojo(name = "setup-sdk", requiresProject = false)
 public class SetupSDK extends AbstractTask {
 
+	@Parameter(defaultValue = "false", property = "enableDataSaving")
+	private boolean enableDataSaving;
```
dkayiwa, line 26: I feel like we should change the above name to something that reflects what exactly the effects of this flag will be on the actual implementation.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -32,19 +36,38 @@ public class NodeHelper {
...
 		this.mavenSession = mavenSession;
 		this.pluginManager = pluginManager;
 		try {
-			this.tempDir = Files.createTempDirectory("openmrs-sdk-node");
-			this.tempDir.toFile().deleteOnExit();
+			Properties sdkProperties = getSdkProperties();
+			boolean isDataSavingMode = Boolean.parseBoolean(sdkProperties.getProperty("enableDataSaving"));
+			String cacheName = "openmrs-sdk-node";
+			if (isDataSavingMode) {
+				logger.info("OpenMRS SDK is running on data saving mode");
```
dkayiwa, line 56: I would not call this a data saving mode because it may be used for other reasons like performance instead of just saving data.

maven-plugin/src/main/java/org/openmrs/maven/plugins/SetupSDK.java
```diff
@@ -13,10 +14,17 @@
...
+import java.util.Properties;
+
+import static org.openmrs.maven.plugins.utility.PropertiesUtils.getSdkProperties;
+import static org.openmrs.maven.plugins.utility.PropertiesUtils.savePropertiesChangesToFile;
 
 @Mojo(name = "setup-sdk", requiresProject = false)
 public class SetupSDK extends AbstractTask {
 
+	@Parameter(defaultValue = "true", property = "removeCacheAfterSetup")
+	private boolean removeCacheAfterSetup;
```
dkayiwa, line 26: Can you add some documentation to this parameter?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -102,5 +125,18 @@ private void runFrontendMavenPlugin(String goal, List<MojoExecutor.Element> conf
 				executionEnvironment(mavenProject, mavenSession, pluginManager)
 		);
 	}
+
+	private File getExistingCache() {
+		String tempDirPath = System.getProperty("java.io.tmpdir");
+		File tempDir = new File(tempDirPath);
+		if(tempDir.exists() && tempDir.isDirectory()) {
+			File[] cache = tempDir.listFiles(file -> file.isDirectory() && file.getName().startsWith("openmrs-sdk-node"));
```
dkayiwa, line 133: https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-CodeStyle

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -102,5 +125,18 @@ private void runFrontendMavenPlugin(String goal, List<MojoExecutor.Element> conf
...
+			if(cache == null) return null;
```
dkayiwa, line 134: https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-CodeStyle

pom.xml
```diff
@@ -433,6 +433,13 @@
                 <artifactId>org.eclipse.jgit</artifactId>
                 <version>4.3.1.201605051710-r</version>
             </dependency>
+
+            <!-- https://mvnrepository.com/artifact/com.google.guava/guava -->
+            <dependency>
+                <groupId>com.google.guava</groupId>
```
dkayiwa, line 439: What are we using this dependency for?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -32,19 +36,38 @@ public class NodeHelper {
...
 		this.mavenSession = mavenSession;
 		this.pluginManager = pluginManager;
 		try {
-			this.tempDir = Files.createTempDirectory("openmrs-sdk-node");
-			this.tempDir.toFile().deleteOnExit();
+			Properties sdkProperties = getSdkProperties();
+			boolean isRemoveCacheAfterSetup = Boolean.parseBoolean(sdkProperties.getProperty("removeCacheAfterSetup"));
+			String cacheName = "openmrs-sdk-node";
+			if (!isRemoveCacheAfterSetup) {
+				logger.info("OpenMRS SDK will reuse the NPM cache. Run openmrs-sdk:setup-sdk -DremoveCacheAfterSetup=true to disable it.");
```
dkayiwa, line 56: Can we be consistent? The description says `reuse npm cache` but the method name says `remove cache`.

pom.xml
wikumChamith: This is the most reliable way I found to delete the NPM cache. Using `org.apache.commons.io.FileUtils` was throwing errors.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
wikumChamith: Should we rename the parameter as "ReuseNpmCache" ??

wikumChamith: I think logger info is consistent with implementation. If isRemoveCacheAfterSetup is false the NPM will reuse NPM caches without deleting them.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -102,5 +125,20 @@ private void runFrontendMavenPlugin(String goal, List<MojoExecutor.Element> conf
...
+
+	private File getExistingCache() {
+		String tempDirPath = System.getProperty("java.io.tmpdir");
+		File tempDir = new File(tempDirPath);
+		if (tempDir.exists() && tempDir.isDirectory()) {
+			File[] cache = tempDir.listFiles(file -> file.isDirectory() && file.getName().startsWith("openmrs-sdk-node"));
+			if (cache == null) {
+				return null;
+			}
+			if (cache.length > 0) {
```
dkayiwa, line 137: What happens to the rest of the caches?

maven-plugin/src/main/java/org/openmrs/maven/plugins/SetupSDK.java
```diff
@@ -13,10 +14,22 @@
...
 
 @Mojo(name = "setup-sdk", requiresProject = false)
 public class SetupSDK extends AbstractTask {
 
+	/**
+	 * Controls whether the SDK removes the NPM cache after the setup.
+	 * By default, this is set to 'true' to save storage space.
+	 * If set to 'false', the SDK will reuse the existing NPM cache
+	 */
+	@Parameter(defaultValue = "true", property = "removeCacheAfterSetup")
```
dkayiwa, line 30: Shouldn't the default be what it was before these changes?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
wikumChamith: Should we delete them??

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
dkayiwa: How about something along the lines of `reuseNodeCache`?

wikumChamith: sounds good

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
dkayiwa: If they are more than one, will the first one be consistently the same for subsequent calls?

wikumChamith: yes

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
wikumChamith: I changed the parameter name

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
wikumChamith: What do you suggest we should do here? We could delete the other caches.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -32,19 +36,38 @@ public class NodeHelper {
...
+	}
+
+	public NodeHelper(MavenProject mavenProject, MavenSession mavenSession, BuildPluginManager pluginManager) {
 		this.mavenProject = mavenProject;
 		this.mavenSession = mavenSession;
 		this.pluginManager = pluginManager;
 		try {
-			this.tempDir = Files.createTempDirectory("openmrs-sdk-node");
-			this.tempDir.toFile().deleteOnExit();
+			Properties sdkProperties = getSdkProperties();
```
ibacher, line 52: It's preferrable, I think to move this code into `getTempDir()` that way we can defer creating a temporary directory until we actually need it.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
```diff
@@ -32,19 +36,38 @@ public class NodeHelper {
...
+			boolean reuseNodeCache = Boolean.parseBoolean(sdkProperties.getProperty("reuseNodeCache"));
+			String cacheName = "openmrs-sdk-node";
```
ibacher, line 54: Seems like this should just be a class constant, no?

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/NodeHelper.java
ibacher: > yes

This actually isn't guaranteed. We're operating on files in the order returned from `listFiles()` which has this caveat:

"There is no guarantee that the name strings in the resulting array will appear in any specific order; they are not, in particular, guaranteed to appear in alphabetical order. "

Similarly, `createTempDirectory()` says:

"The details as to how the name of the directory is constructed is implementation dependent and therefore not specified."

In order to guarantee that this always returns the same folder, we'd need a guarantee that `listFiles()` returned files in a consistent order and that `createTempDirectory()` generates files in a defined order that will fit the order returned by `listFiles()`, but those details are specifically disclaimed.

(On Windows, the default JVM does indeed use the tick count as the source of randomness for `createTempDirectory()` and `listFiles()` likely returns in a consistent order, which would likely result in this appearing to work; however, Linux, the underlying `mkstemp()` implementation will try to use a RNG, falling back to ticks if not available on that device. I haven't looked into the details of macOS.)

Probably the thing to do is to make the `tempDir` a class variable instead of an instance variable so it's only set once per JVM, which would guarantee consistent results across multiple invocations of the NodeHelper within a single invocation of the SDK.

wikumChamith:
@ibacher I did the changes you requested.

ibacher:
@wikumChamith I think what I actually asked for was this:

> On reflection, actually, I think a better strategy is: if the user wants to use a consistent NPM cache directory, use a named folder in the directory the SDK "servers directory" (defined [here](https://github.com/openmrs/openmrs-sdk/blob/815314d8e161f8ea0af9b712412b14e4edd0b31a/sdk-commons/src/main/java/org/openmrs/maven/plugins/model/Server.java#L73)). Just name it something unlikely to be a server name, e.g., _openmrs_sdk_node_cache. That's cleaner, easier for end-users to understand and nobody has to go poking around for the right "openmrs-sdk-node.1234567" folder.

Also, even if we have a global option for this (which is fine), it would be nice to have a local option to override the global option without needing to run `setup-sdk` all over again (especially as that has side-effects I'd rather avoid).

wikumChamith:
@ibacher I  made the changes you suggested.

ibacher (commented):
Thanks @wikumChamith! There are two more places we should be able to override the node cache value:

1. BuildDistro
2. ServerUpgrader

Then I think this will be all set. Thanks for the work on this!

wikumChamith:
@ibacher I integrated the cache management to build-distro and server upgrader.

ibacher (approved):

makombe:
S3yg

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
```diff
@@ -33,6 +36,7 @@
 public class PropertiesUtils {
 
 	private static final Logger log = LoggerFactory.getLogger(PropertiesUtils.class);
+	private static final String SDK_PROPERTIES_FILE = "SDK Properties file";
```
ibacher, line 39: Shouldn't this just be something like `sdk.properties`?

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
```diff
@@ -234,6 +238,40 @@ public static Properties getModuleProperty(String url) {
...
+	 *
+	 * @param properties The {@link Properties} to be saved to the file.
+	 * @param file       The {@link File} where the properties will be saved.
+	 * @throws MojoExecutionException If an {@link IOException} occurs during the file writing process,
+	 *                               or if there is an issue with the provided properties or file.
+	 */
+	public static void savePropertiesChangesToFile(Properties properties, File file)
+			throws MojoExecutionException {
+		try (OutputStream fos = new FileOutputStream(file)) {
+			properties.store(fos, SDK_PROPERTIES_FILE + ":");
```
dkayiwa, line 267: What is the use of the full colon at the end of this comment?

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
wikumChamith: When writing the SDK properties to the files using [Properties#store(java.io.OutputStream, java.lang.String)](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#store-java.io.OutputStream-java.lang.String-) we have to pass a comment. This comment is used as the heading of the file. I renamed the variable name to eliminate any confusion. The actual file name of the properties file is "sdk.properties".

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
wikumChamith: I removed the colon.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
dkayiwa: Then the constant name would be less confusing if it reflected that in one way or another like SDK_PROPERTIES_FILE_HEADER. Or just get rid of the constant because after all, you use it in only one place.

wikumChamith: @dkayiwa I removed the variable in this PR: https://github.com/openmrs/openmrs-sdk/pull/257

PR #279, opened 2024-07-31
O3-3426: Distro Inheritance Tool for the Build-Distro plugin

wikumChamith:

Adding distro inheritance support to the build-distro plugin.

## Issue I worked on

see https://openmrs.atlassian.net/browse/O3-3431

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -138,6 +138,8 @@ public class BuildDistro extends AbstractTask {
 			if (distroFile.exists()) {
 				wizard.showMessage("Building distribution from the distro file at " + distroFile + "...\n");
 				distroProperties = new DistroProperties(distroFile);
+				distroArtifact = distroProperties.getParentArtifact();
+				distroProperties = handleParentDistroProperties(distroArtifact, buildDirectory, distroProperties);
 			} else if (Project.hasProject(userDir)) {
 				Project config = Project.loadProject(userDir);
 				distroArtifact = DistroHelper
@@ -157,6 +159,8 @@ public class BuildDistro extends AbstractTask {
 			}
 		} else if (StringUtils.isNotBlank(distro)) {
 			distroProperties = distroHelper.resolveDistroPropertiesForStringSpecifier(distro, versionsHelper);
+			distroArtifact = distroProperties.getParentArtifact();
+			distroProperties = handleParentDistroProperties(distroArtifact, buildDirectory, distroProperties);
 		}
 
 		if (distroProperties == null) {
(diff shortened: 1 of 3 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -198,6 +202,18 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
 						+ buildDirectory.getAbsolutePath() + "\n");
 	}
 
+	private DistroProperties handleParentDistroProperties(Artifact distroArtifact, File buildDirectory, DistroProperties distroProperties) throws MojoExecutionException {
+		if (StringUtils.isNotBlank(distroArtifact.getArtifactId()) && StringUtils.isNotBlank(distroArtifact.getGroupId()) && StringUtils.isNotBlank(distroArtifact.getVersion())) {
+			Server tempServer = new Server.ServerBuilder().build();
```
ibacher, line 207: Is there a way we could de-couple this from the server setup? It seems like the only real reason we pass this in is for the server directory so maybe `resolveParentArtifact()` and `getArtifactProperties()` could be updated to just take a directory as an argument?

ibacher (commented):
Is the code here recursive, i.e., if my parent distro has a parent is that also processed?

wikumChamith:
> Is the code here recursive, i.e., if my parent distro has a parent is that also processed?

@ibacher, this implementation isn't recursive. Once it's in place, how can I test it by creating a parent distro that also has its parent?

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
wikumChamith: I updated the PR with these changes.

ibacher (approved):

ibacher:
> Once it's in place, how can I test it by creating a parent distro that also has its parent?

Hmmm... I'm not entirely sure. I suppose if we have / added a way to resolve to a local directory rather than via Maven that might work? Alternatively, we can create dummy distro artifacts...

wikumChamith:
> > Once it's in place, how can I test it by creating a parent distro that also has its parent?
> 
> Hmmm... I'm not entirely sure. I suppose if we have / added a way to resolve to a local directory rather than via Maven that might work? Alternatively, we can create dummy distro artifacts...

@ibacher can you provide more information on how to do that?

ibacher:
I think it would be something like the bundled modules we already have for the "build" integration tests. Basically, we'd have some "bundled" dummy distributions parent1 and parent2 where parent2 depends on parent1. Then, before running a test for them, we use Maven to install those distributions in the local repo, and hopefully the tooling should just work.

wikumChamith:
> I think it would be something like the bundled modules we already have for the "build" integration tests. Basically, we'd have some "bundled" dummy distributions parent1 and parent2 where parent2 depends on parent1. Then, before running a test for them, we use Maven to install those distributions in the local repo, and hopefully the tooling should just work.

Thank you. I'll merge this and move on to that in a separate PR.

PR #280, opened 2024-07-31
O3-3512: SDK should handle reading dependencies from content.properties and ensuring they are part of the distro

mherman22:
## Issue I worked on

see https://openmrs.atlassian.net/browse/O3-3512

- [ ] All new and existing **tests passed**.

### DIFF
--- a/integration-tests/src/test/resources/integration-test/openmrs-distro.properties
+++ b/integration-tests/src/test/resources/integration-test/openmrs-distro.properties
@@ -13,3 +13,4 @@ spa.frontendModules.@openmrs/esm-login-app=3.3.1
 spa.frontendModules.@openmrs/esm-patient-chart-app=3.1.0
 spa.apiUrl=notopenmrs
 spa.configUrls=foo
+content.hiv=1.0.0
--- a/maven-plugin/pom.xml
+++ b/maven-plugin/pom.xml
@@ -30,11 +30,6 @@
             <artifactId>maven-verifier</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.vdurmont</groupId>
-            <artifactId>semver4j</artifactId>
-            <version>3.1.0</version>
-        </dependency>
-        <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-model</artifactId>
         </dependency>
--- a/maven-plugin/pom.xml
+++ b/maven-plugin/pom.xml
@@ -30,6 +30,11 @@
             <artifactId>maven-verifier</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.vdurmont</groupId>
+            <artifactId>semver4j</artifactId>
+            <version>3.1.0</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-model</artifactId>
         </dependency>
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -1,6 +1,5 @@
 package org.openmrs.maven.plugins;
 
-import com.github.zafarkhaja.semver.ParseException;
 import net.lingala.zip4j.core.ZipFile;
 import net.lingala.zip4j.exception.ZipException;
 import net.lingala.zip4j.model.ZipParameters;
@@ -26,8 +25,6 @@ import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileWriter;
-import java.io.BufferedReader;
-import java.io.FileReader;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -36,7 +33,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
-import java.util.Properties;
 
 /**
  * Create docker configuration for distributions.
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -12,10 +12,10 @@ import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.openmrs.maven.plugins.model.Artifact;
 import org.openmrs.maven.plugins.model.DistroProperties;
+import org.openmrs.maven.plugins.model.Project;
 import org.openmrs.maven.plugins.model.Server;
 import org.openmrs.maven.plugins.model.Version;
 import org.openmrs.maven.plugins.utility.DistroHelper;
-import org.openmrs.maven.plugins.model.Project;
 import org.openmrs.maven.plugins.utility.PropertiesUtils;
 import org.openmrs.maven.plugins.utility.SDKConstants;
 import org.slf4j.Logger;
@@ -193,7 +193,7 @@ public class BuildDistro extends AbstractTask {
 			throw new MojoExecutionException("The distro you specified, '" + distro + "' could not be retrieved");
 		}
 
-		PropertiesUtils.parseContentProperties(userDir, distroProperties);
+		PropertiesUtils.parseContentProperties(distroProperties);
 		String distroName = buildDistro(buildDirectory, distroArtifact, distroProperties);
 
 		wizard.showMessage(
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -1,5 +1,7 @@
 package org.openmrs.maven.plugins;
 
+import com.vdurmont.semver4j.Semver;
+import com.vdurmont.semver4j.SemverException;
 import net.lingala.zip4j.core.ZipFile;
 import net.lingala.zip4j.exception.ZipException;
 import net.lingala.zip4j.model.ZipParameters;
@@ -8,11 +10,8 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
-import org.eclipse.jgit.api.errors.GitAPIException;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
-import org.eclipse.jgit.api.CloneCommand;
-import org.eclipse.jgit.api.Git;
 import org.openmrs.maven.plugins.model.Artifact;
 import org.openmrs.maven.plugins.model.DistroProperties;
 import org.openmrs.maven.plugins.model.Server;
@@ -27,6 +26,8 @@ import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileWriter;
+import java.io.BufferedReader;
+import java.io.FileReader;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -35,6 +36,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.Properties;
 
 /**
  * Create docker configuration for distributions.
@@ -47,6 +49,8 @@ public class BuildDistro extends AbstractTask {
 	private static final String OPENMRS_WAR = "openmrs.war";
 
 	private static final String OPENMRS_DISTRO_PROPERTIES = "openmrs-distro.properties";
+
+	private static final String CONTENT_PROPERTIES = "content.properties";
 
 	private static final String DOCKER_COMPOSE_PATH = "build-distro/docker-compose.yml";
 
@@ -78,7 +82,7 @@ public class BuildDistro extends AbstractTask {
 
 	private static final Logger log = LoggerFactory.getLogger(BuildDistro.class);
 
-	/**
+    /**
 	 * Path to the openmrs-distro.properties file.
 	 */
 	@Parameter(property = "distro")
@@ -195,6 +199,7 @@ public class BuildDistro extends AbstractTask {
 			throw new MojoExecutionException("The distro you specified, '" + distro + "' could not be retrieved");
 		}
 
+		parseContentProperties(userDir, distroProperties);
 		String distroName = buildDistro(buildDirectory, distroArtifact, distroProperties);
 
 		wizard.showMessage(
@@ -510,7 +515,7 @@ public class BuildDistro extends AbstractTask {
 		}
 
 		try (FileWriter writer = new FileWriter(dbDump);
-		     BufferedInputStream bis = new BufferedInputStream(stream)) {
+			 BufferedInputStream bis = new BufferedInputStream(stream)) {
 			writer.write(DUMP_PREFIX);
 
 			int c;
@@ -603,5 +608,88 @@ public class BuildDistro extends AbstractTask {
 				throw new MojoExecutionException("Distro should contain only single war file");
 			}
 		}
+	}
+
+	/**
+	 * Reads the content.properties file.
+	 * <p> For dependencies defined in both content.properties and distro.properties, this ensures the version specified in distro.properties 
+	 * matches the range defined in content.properties; otherwise, we output an error to the user to fix this.</p>
+	 * For dependencies not defined in distro.properties but defined in content.properties, the SDK should locate the latest matching version and add that to the OMODs and ESMs used by the distro.
+	 * 
+	 * @param userDir
+	 * @param distroProperties
+	 * @throws MojoExecutionException
+	 */
+	private void parseContentProperties(File userDir, DistroProperties distroProperties) throws MojoExecutionException {
+		File contentFile = new File(userDir, CONTENT_PROPERTIES);
+		if (!contentFile.exists()) {
+			log.info("No content.properties file found in the user directory.");
+			return;
+		}
+
+		Properties contentProperties = new Properties();
+		try (BufferedReader reader = new BufferedReader(new FileReader(contentFile))) {
+			contentProperties.load(reader);
+		} catch (IOException e) {
+			throw new MojoExecutionException("Failed to read content.properties file", e);
+		}
+
+		for (String dependency : contentProperties.stringPropertyNames()) {
+			String versionRange = contentProperties.getProperty(dependency);
+			String distroVersion = distroProperties.get(dependency);
+
+			if (distroVersion == null) {
+				String latestVersion = findLatestMatchingVersion(dependency, versionRange);
+				if (latestVersion == null) {
+					throw new MojoExecutionException("No matching version found for dependency " + dependency);
+				}
+
+				distroProperties.add(dependency, latestVersion);
+			} else {
+				checkVersionInRange(dependency, versionRange, distroVersion);
+			}
+		}
+	}
+
+	/**
+     * Checks if the version from distro.properties satisfies the range specified in content.properties.
+     * Throws an exception if there is a mismatch.
+     *
+     * @param contentDependencyKey       the dependency key
+     * @param contentDependencyVersionRange the version range from content.properties
+     * @param distroPropertyVersion      the version from distro.properties
+     * @throws MojoExecutionException if there is a version mismatch
+     */
+	private void checkVersionInRange(String contentDependencyKey, String contentDependencyVersionRange, String distroPropertyVersion) throws MojoExecutionException {
+		Semver semverVersion = new Semver(distroPropertyVersion, Semver.SemverType.NPM);
+		String[] ranges = contentDependencyVersionRange.split(",");
+		boolean inRange = false;
+
+		for (String range : ranges) {
+			range = range.trim();
+			try {
+				if (range.startsWith("^") || range.startsWith("~")) {
+					inRange = semverVersion.satisfies(range);
+				} else {
+					Semver rangeVersion = new Semver(range, Semver.SemverType.NPM);
+					inRange = semverVersion.diff(rangeVersion) == Semver.VersionDiff.NONE;
+				}
+				if (inRange) break;
+			} catch (SemverException e) {
+				throw new MojoExecutionException("Invalid version range format for " + contentDependencyKey + ": " + contentDependencyVersionRange, e);
+			}
+		}
+
+		if (!inRange) {
+			throw new MojoExecutionException(
+					"Incompatible version for " + contentDependencyKey + ". Specified range: " + contentDependencyVersionRange
+							+ ", found in distribution: " + distroPropertyVersion);
+		}
+	}
+
+	private String findLatestMatchingVersion(String dependency, String versionRange) {
+		// TODO
+		// Implement logic to find the latest version matching the versionRange
+		return null;
 	}
 }
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -16,7 +16,6 @@ import org.openmrs.maven.plugins.model.Project;
 import org.openmrs.maven.plugins.model.Server;
 import org.openmrs.maven.plugins.model.Version;
 import org.openmrs.maven.plugins.utility.DistroHelper;
-import org.openmrs.maven.plugins.utility.PropertiesUtils;
 import org.openmrs.maven.plugins.utility.SDKConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -193,7 +192,7 @@ public class BuildDistro extends AbstractTask {
 			throw new MojoExecutionException("The distro you specified, '" + distro + "' could not be retrieved");
 		}
 
-		PropertiesUtils.parseContentProperties(distroProperties);
+		distroHelper.parseContentProperties(distroProperties);
 		String distroName = buildDistro(buildDirectory, distroArtifact, distroProperties);
 
 		wizard.showMessage(
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
@@ -651,6 +651,10 @@ public class BuildDistro extends AbstractTask {
 		}
 	}
 
+	private String findLatestMatchingVersion(String dependency, String versionRange) throws MojoExecutionException {
+		return distroHelper.findLatestMatchingVersion(dependency, versionRange);
+	}
+
 	/**
      * Checks if the version from distro.properties satisfies the range specified in content.properties.
      * Throws an exception if there is a mismatch.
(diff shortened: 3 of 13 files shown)


maven-plugin/pom.xml
```diff
@@ -29,6 +29,11 @@
             <groupId>org.apache.maven.shared</groupId>
             <artifactId>maven-verifier</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.vdurmont</groupId>
+            <artifactId>semver4j</artifactId>
+            <version>3.1.0</version>
+        </dependency>
```
ibacher, line 36: Could this be implemented using the java-semver library we already have instead of adding a new dependency?

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -587,4 +592,85 @@ public boolean accept(File dir, String name) {
...
+	 * Reads the content.properties file.
+	 * <p> For dependencies defined in both content.properties and distro.properties, this ensures the version specified in distro.properties 
+	 * matches the range defined in content.properties; otherwise, we output an error to the user to fix this.</p>
+	 * For dependencies not defined in distro.properties but defined in content.properties, the SDK should locate the latest matching version and add that to the OMODs and ESMs used by the distro.
+	 * 
+	 * @param userDir
+	 * @param distroProperties
+	 * @throws MojoExecutionException
+	 */
+	private void parseContentProperties(File userDir, DistroProperties distroProperties) throws MojoExecutionException {
```
ibacher, line 606: These checks should be implemented for both the `build-distro` and `setup` commands, so it might make sense to extract the "meat" of them into one of the shared components (or add a new one).

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -587,4 +592,85 @@ public boolean accept(File dir, String name) {
...
+				String latestVersion = findLatestMatchingVersion(dependency, versionRange);
+				if (latestVersion == null) {
+					throw new MojoExecutionException("No matching version found for dependency " + dependency);
+				}
+
+				distroProperties.add(dependency, latestVersion);
+			} else {
+				checkVersionInRange(dependency, versionRange, distroVersion);
+			}
+		}
```
ibacher, line 634: Not all properties in the `content.properties` specify some sort of library. Specifically, I think it's only the keys that start: `omod.`, `owa.`, and `spa.frontendModule.` (plus, I suppose, `content.`

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -587,4 +592,85 @@ public boolean accept(File dir, String name) {
...
+	}
+
+	private String findLatestMatchingVersion(String dependency, String versionRange) throws MojoExecutionException {
+		return distroHelper.findLatestMatchingVersion(dependency, versionRange);
```
ibacher, line 638: Here we need slightly different handling for `findLatestMatchingVersion` for keys that are `spa.frontendModule.` and others, since frontendModule versions need to be looked up from an NPM registry rather than a Maven registry.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -587,4 +592,85 @@ public boolean accept(File dir, String name) {
...
+     * Throws an exception if there is a mismatch.
+     *
+     * @param contentDependencyKey       the dependency key
+     * @param contentDependencyVersionRange the version range from content.properties
+     * @param distroPropertyVersion      the version from distro.properties
+     * @throws MojoExecutionException if there is a version mismatch
+     */
+	private void checkVersionInRange(String contentDependencyKey, String contentDependencyVersionRange, String distroPropertyVersion) throws MojoExecutionException {
+		Semver semverVersion = new Semver(distroPropertyVersion, Semver.SemverType.NPM);
+		String[] ranges = contentDependencyVersionRange.split(",");
```
ibacher, line 652: I don't think we need to support comma-separated ranges. The default SemVer ranges have enough disjuctivity.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -587,4 +592,85 @@ public boolean accept(File dir, String name) {
...
+		boolean inRange = false;
+
+		for (String range : ranges) {
+			range = range.trim();
+			try {
+				if (range.startsWith("^") || range.startsWith("~")) {
+					inRange = semverVersion.satisfies(range);
+				} else {
```
ibacher, line 660: I think the test here should always be something like `semverVersion.satisfies(range)`. Ranges can also be specified as things like `1.x` or `1.10.1 || ^2 || ~4.0 || > 5`.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -587,4 +592,85 @@ public boolean accept(File dir, String name) {
...
+					inRange = semverVersion.diff(rangeVersion) == Semver.VersionDiff.NONE;
+				}
+				if (inRange) break;
+			} catch (SemverException e) {
+				throw new MojoExecutionException("Invalid version range format for " + contentDependencyKey + ": " + contentDependencyVersionRange, e);
+			}
+		}
+
+		if (!inRange) {
+			throw new MojoExecutionException(
```
ibacher, line 671: Error message should probably also include a way of identifying the content package in question since a distro may have several.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
ibacher: Oh! and `war`!

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -517,4 +518,11 @@ public Properties getArtifactProperties(Artifact artifact, Server server, String
 		return properties;
 	}
 
+	public String findLatestMatchingVersion(String dependency, String versionRange) throws MojoExecutionException {
+		if (dependency.startsWith("omod") || dependency.startsWith("owa") || dependency.startsWith("spa.frontend")) {
```
ibacher, line 522: Should probably support `war` too, no?

ibacher (changes requested):
Thanks @mherman22! Great work here. A few things that need to be fixed, but a very good first pass. Also note the `O` in `O3` is the letter "O" rather than the number `0`.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
mherman22: I agree

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
mherman22: Oh, makes perfect sense

maven-plugin/pom.xml
mherman22: True, I've just seen that, lemi use the existing library!

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
mherman22: @ibacher For example, what will be the value for `content.`?

ibacher: `content.hiv=1.0.0` or something. If we have a content package that requires another content package.

ibacher: (Assume the default groupId is `org.openmrs.content`, but can be anything using the `content.hiv.groupId=xxx` pattern).

mherman22: ooh perfect!

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/NpmVersionHelper.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import org.json.JSONObject;
+
+public class NpmVersionHelper {
+
+    public String getLatestReleasedVersionFromNpmRegistry(PackageJson packageJson) {
```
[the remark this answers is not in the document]
ibacher: That said, the overall goal here is that the user supplies some semver range and we resolve that to a version which may or may not always be the latest version. Here's what I'd suggest: use the existing NPM helper stuff to run `npm pack --dry-run --json <package>@<version>`, e.g., `npm pack --dry-run --json '@openmrs/esm-login-app@>=5.0.0-pre.1'` will get you a JSON document that has the actual resolved version, etc.

ibacher (commented):
Quick remark. I haven't fully reviewed. Also, based on today's conversation, can we switch from the java-semver library to semver4j. Sorry for the whiplash!

maven-plugin/pom.xml
mherman22: done!

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
```diff
@@ -301,4 +309,87 @@ private static Document parseXMLFromURL(String url) throws ParserConfigurationEx
...
+	 * matches the range defined in content.properties; otherwise, we output an error to the user to fix this.</p>
+	 * For dependencies not defined in distro.properties but defined in content.properties, the SDK should locate the latest matching version
+	 * and add that to the OMODs and ESMs used by the distro.
+	 *
+	 * @param userDir           The directory where the distro.properties or content.properties file is located. This is typically the user's working directory.
+	 * @param distroProperties  An object representing the properties defined in distro.properties. This includes versions of various dependencies
+	 *                          currently used in the distribution.
+	 * @throws MojoExecutionException if there is an error reading the content.properties file or if a version conflict is detected.
+	 */
+	public static void parseContentProperties(File userDir, DistroProperties distroProperties) throws MojoExecutionException {
```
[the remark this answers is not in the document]
mherman22: i am struggling to get what you mean here @ibacher.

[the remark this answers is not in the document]
ibacher: Yeah... past Ian doesn't make a lot of sense. I _think_ what I was getting at is that this assumes that the `content.properties` file is in the directory with all the other files, but it's likely inside a _zip file_ inside that folder or something.

I also couldn't quite see how things would work if there is more than one content package (which is a use-case we're supporting)?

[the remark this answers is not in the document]
mherman22: Thanks , however I thought this particular ticket was to basically pick a content.properties that has already been extracted by https://openmrs.atlassian.net/browse/O3-3640 and then do the rest of the stuff?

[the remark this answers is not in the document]
mherman22: Cc: @ibacher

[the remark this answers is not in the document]
mherman22: Otherwise if I'm wrong,  feel free to review the latest commit 😄

sdk-commons/pom.xml
```diff
@@ -81,6 +86,11 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>20230618</version>
+        </dependency>
```
ibacher, line 93: Forgot this: we actually have Jackson as a dependency already. Can we convert the code to use that instead of adding a new JSON-handling library please?

sdk-commons/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
```diff
@@ -291,4 +291,30 @@ private boolean hasPlaceholder(Object object){
...
+    public void add(String dependency, String version) {
+        if (StringUtils.isBlank(dependency) || StringUtils.isBlank(version)) {
+            log.error("Dependency name or version cannot be blank");
+            return;
+        }
+
+        if (dependency.startsWith("omod") || dependency.startsWith("war")) {
+            properties.setProperty(dependency, version);
+            log.info("Added dependency: {} with version: {}", dependency, version);
+        }
```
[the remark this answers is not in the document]
mherman22: true!

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/NpmVersionHelper.java
```diff
@@ -0,0 +1,68 @@
...
+                .inheritIO();
+        Process process = processBuilder.start();
+
+        // Read the command output
+        BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
+        StringBuilder outputBuilder = new StringBuilder();
+        String line;
+        while ((line = reader.readLine()) != null) {
+            outputBuilder.append(line);
+        }
```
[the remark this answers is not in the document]
mherman22: thanks, though im gonna use a `Reader` than an `InputStream`. It is really a byte vs char war

sdk-commons/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
```diff
@@ -291,4 +291,31 @@ private boolean hasPlaceholder(Object object){
...
+		        || dependency.startsWith("spa.frontendModule")) {
+			properties.setProperty(dependency, version);
+			log.info("Added dependency: {} with version: {}", dependency, version);
+		}
+	}
+	
+	public String get(String contentDependencyKey) {
+		return properties.getProperty(contentDependencyKey);
+	}
 }
```
ibacher, line 321: This file is mostly indented with spaces, but this change adds tabs for this section only. It's preferable to be consistent about this.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -48,14 +50,18 @@ public class DistroHelper {
...
 	public DistroHelper(MavenProject mavenProject, MavenSession mavenSession, BuildPluginManager pluginManager,
-						Wizard wizard, VersionsHelper versionHelper) {
+                        Wizard wizard, VersionsHelper versionHelper) {
 		this.mavenProject = mavenProject;
 		this.mavenSession = mavenSession;
 		this.pluginManager = pluginManager;
 		this.wizard = wizard;
 		this.versionHelper = versionHelper;
-	}
+    }
```
ibacher, line 64: More tabs v spaces changes.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -48,14 +50,18 @@ public class DistroHelper {
 
 	final VersionsHelper versionHelper;
 
+	NpmVersionHelper npmVersionHelper;
```
ibacher, line 53: Does this ever get set?

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -48,14 +50,18 @@ public class DistroHelper {
...
+
+	PackageJson packageJson;
```
ibacher, line 55: Do we ever use this?

mherman22: yes we do! within the same class at:
```
	public String findLatestMatchingVersion(String dependency, String versionRange) {
		if (dependency.startsWith("omod") || dependency.startsWith("owa") || dependency.startsWith("content.") || dependency.startsWith("war.")) {
			return versionHelper.getLatestReleasedVersion(new Artifact(dependency, "latest"));
		} else if (dependency.startsWith("spa.frontendModule")) {
			packageJson.setName(dependency.substring("spa.frontendModules.".length()));
			return npmVersionHelper.getResolvedVersionFromNpmRegistry(packageJson, versionRange);
		}
		throw new IllegalArgumentException("Unsupported dependency type: " + dependency);
	}
```

ibacher: Ah! Where does it get initialized?

mherman22:
> We may need to think about how to get the content packages Zip. I'm not entirely sure that we store the Zip file anywhere other than in the Maven repo. Maybe we use the Maven utilities to move the content packages to a [temporary directory](https://www.baeldung.com/java-temp-directories#delegate-to-the-operating-system)?

is this the implementation that is expected? i.e we have a `distro.properties` file which has all the properties and their corresponding values needed to build a distro and it looks something like, 
```
name=Ref 3.x 
distro version=3.0.0
war.openmrs=2.7.0-SNAPSHOT
omod.fhir2=1.9.0
omod.webservices.rest=2.39.0
omod.uicommons=2.23.0
omod.uiframework=3.23.0
omid.uilibrary=2.0.7
omod.idgen=4.10.0
omod.spa=1.0.6
spa.core=3.3.1
spa.frontendModules.@openmrs/esm-login-app=3.3.1 
spa.frontendModules.@openmrs/esm-patient-chart-app=3.1.0
content.hiv=1.0.0
content.pmtc=2.0.0
content.anc=2.4.5
```
Should be able to filter out the properties starting with `content.` and fetch those versions of each content property key from the maven **repo**, store them in a temporary directory that is created and should be deleted after the processing.

Now go into the temp directory and process/unzip the zip files in there by also making sure each zip contains at least one `content.properties` file.

And then we process the `content.properties` by reading all underlying `content.properties` files and check dependencies against `distro.properties`. 

For dependencies defined in both `content.properties` and `distro.properties`, this should ensure that the version specified in `distro.properties` matches the range defined in `content.properties`; Otherwise, it outputs an error for the user to fix.

For dependencies not defined in `distro.properties` but defined in content.properties, the SDK should locate the latest matching version and add that to the OMODs and ESMs used by the distro.

My latest commit tries to include all the above, still looking into fetching from maven for now.

ibacher:
Sounds about right. Now that I think about it, I think we can do something very similar to the `[downloadDistroProperties()`](https://github.com/openmrs/openmrs-sdk/blob/0285a8c64425246b647e65601d032b272c32e8c3/sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java#L253C26-L253C50) stuff, right? That:

* Locates a Zip file using Maven
* Downloads it to a location
* Extracts the distro.properties

> still looking into fetching from maven for now

Both the ModuleInstaller and the DistroHelper have similar code paths that download Maven artifacts and copy them to appropriate locations.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
```diff
@@ -301,4 +314,184 @@ private static Document parseXMLFromURL(String url) throws ParserConfigurationEx
...
+		try {
+			tempDirectory = Files.createTempDirectory("content-packages").toFile();
+			String[] contentPackages = getContentPackages(distroProperties);
+
+			for (String contentPackage : contentPackages) {
+				String version = distroProperties.get(contentPackage);
+				File zipFile = fetchZipFromMavenRepo(contentPackage, version, tempDirectory);
+				if (zipFile != null) {
+					processZipFile(zipFile, distroProperties);
+				}
```
ibacher, line 340: Wouldn't it be more efficient to just do this:

```suggestion
			for (String key : distroProperties.getAllKeys()) {
				if (!key.startsWith(CONTENT_PREFIX)) {
					continue;
				}
```

Instead of basically iterating things twice? Also, it might make sense to add `getContentArtifacts()` to `BaseSdkProperties` that matches, e.g., `getModuleArtifacts()`?

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
```diff
@@ -301,4 +314,184 @@ private static Document parseXMLFromURL(String url) throws ParserConfigurationEx
...
+				} catch (IOException e) {
+					log.warn("Failed to delete temporary directory: {}", tempDirectory.getAbsolutePath(), e);
+				}
+			}
+		}
+	}
+
+	private static File fetchZipFromMavenRepo(String contentPackage, String version, File tempDirectory) throws MojoExecutionException {
+		// Base URL for the OpenMRS JFrog repository
+		String repositoryUrl = "https://openmrs.jfrog.io/artifactory/public/";
```
ibacher, line 360: Maybe refactor `DistroHelper` and `ModuleInstaller` into a `MavenHelper` to do this? We don't want to assume everything is available in the JFrog repo (there shouldn't be any references to it in the SDK; use `mavenrepo.openmrs.org` in preference to it)

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
```diff
@@ -301,4 +314,184 @@ private static Document parseXMLFromURL(String url) throws ParserConfigurationEx
...
+		String groupId = "org.openmrs.content";
```
ibacher, line 361: There's a way to override groupIds that needs to be supported here too. IIRC the syntax is something like:

```
omod.event=1.0.0
omod.event.groupId=org.openmrs
```

So we'd need to support something like:

```
content.hiv=1.0.0
content.hiv.groupId=com.ozone-his
```

ibacher: Code I'm thinking of is [here](https://github.com/openmrs/openmrs-sdk/blob/master/sdk-commons/src/main/java/org/openmrs/maven/plugins/model/BaseSdkProperties.java#L169).

ibacher (commented):
Nice work! A few more pointers here, but this is really shaping up!

integration-tests/src/test/resources/integration-test/openmrs-distro.properties
```diff
@@ -13,3 +13,4 @@ spa.frontendModules.@openmrs/esm-login-app=3.3.1
 spa.frontendModules.@openmrs/esm-patient-chart-app=3.1.0
 spa.apiUrl=notopenmrs
 spa.configUrls=foo
+content.hiv=1.0.0
```
mherman22, line 16: @ibacher i added this so as to test whether [this test](https://github.com/openmrs/openmrs-sdk/blob/0285a8c64425246b647e65601d032b272c32e8c3/integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIntegrationTest.java#L116) would pass and i ran it locally and it is passing. It is what i used to test the functionality on the `setup` command.

mherman22 (commented):
I moved stuff to DistroHelper because of the so many changes i would have to make to get rid of the `NullPointerException` i was encountering.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -78,7 +76,7 @@ public class BuildDistro extends AbstractTask {
 
 	private static final Logger log = LoggerFactory.getLogger(BuildDistro.class);
 
-	/**
+    /**
```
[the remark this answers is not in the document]
ibacher: Could you double-check that this is a tab and not spaces?

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -553,4 +578,233 @@ public DistroProperties resolveParentArtifact(Artifact parentArtifact, Server se
...
+	}
+
+	/**
+	 * Checks if the groupId for a given dependency is overwritten in the distro properties.
+	 *
+	 * @param dependencyKey    The key of the dependency.
+	 * @param distroProperties The distro properties file.
+	 * @return The groupId to use for this dependency.
+	 */
+	private static String checkIfOverwritten(String dependencyKey, DistroProperties distroProperties) {
```
ibacher, line 758: Is there a reason we can't use the default `checkIfOverwritten()`?

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/NpmVersionHelper.java
```diff
@@ -0,0 +1,76 @@
...
+		
+		ProcessBuilder processBuilder = new ProcessBuilder()
+		        .command("npm", "pack", "--dry-run", "--json", packageName + "@" + versionRange).redirectErrorStream(true)
+		        .inheritIO();
+		Process process = processBuilder.start();
+		
+		// Read the command output
+		StringBuilder outputBuilder = new StringBuilder();
+		char[] buffer = new char[4096];
+		try (Reader reader = new InputStreamReader(process.getInputStream(), StandardCharsets.UTF_8)) {
```
ibacher, line 61: Why are we reading the input stream? Shouldn't this be the output stream?

ibacher (changes requested):
Thanks @mherman22. Have you tested this code out?

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
mherman22: updated!

sdk-commons/src/main/java/org/openmrs/maven/plugins/model/BaseSdkProperties.java
```diff
@@ -167,7 +168,7 @@ protected String getArtifactType(String key){
     }
 
     protected String checkIfOverwritten(String key, String param) {
-        String newKey = key + "." + param;
+        String newKey = key + "." + param; //content.hiv
```
ibacher, line 171: Probably don't need this long-term.

sdk-commons/src/main/java/org/openmrs/maven/plugins/model/BaseSdkProperties.java
```diff
@@ -189,6 +190,8 @@ protected String checkIfOverwritten(String key, String param) {
                         case TYPE_DISTRO:
 	                    case TYPE_CONFIG:
                             return properties.getProperty(PROPERTY_DISTRO_GROUP_ID, Artifact.GROUP_DISTRO);
+                        case TYPE_CONTENT:
+                            return properties.getProperty(param + ".groupId",Artifact.GROUP_CONTENT);
```
ibacher, line 194: Can we get this to a closer parallel to `TYPE_OMOD`? Here we should only need to handle the default value.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/PropertiesUtils.java
```diff
@@ -17,24 +33,7 @@
...
-import org.apache.maven.plugin.MojoExecutionException;
-import org.openmrs.maven.plugins.model.Artifact;
-import org.openmrs.maven.plugins.model.Server;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.w3c.dom.Document;
-import org.xml.sax.SAXException;
-
-public class PropertiesUtils {
+public class PropertiesUtils  {
```
ibacher, line 36: Can we remove the changes from this file, since they are now all not substantive changes?

ibacher:
It looks like the CI build is failing with a NPE here: https://github.com/openmrs/openmrs-sdk/pull/280/files#diff-c64d06786c01f8050b86ec442ebad27e09a108f0f546875e26dc55a3ce57a334R643

mherman22:
> It looks like the CI build is failing with a NPE here: https://github.com/openmrs/openmrs-sdk/pull/280/files#diff-c64d06786c01f8050b86ec442ebad27e09a108f0f546875e26dc55a3ce57a334R643

Looking into it

ibacher (approved):

PR #301, opened 2024-10-30
SDK-306 - Fix to enable the config-prefixed distro properties to succ…

mseaton:
…essfully install config projects outside of the refapp

see https://openmrs.atlassian.net/browse/SDK-306

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
@@ -371,7 +371,7 @@ public class Setup extends AbstractServerTask {
 			// If such a folder is found in the downloaded artifact, use it.  Otherwise, use the entire zip contents
 			File tempConfigDir = new File(server.getServerTmpDirectory(), UUID.randomUUID().toString());
 			try {
-				if (!tempConfigDir.mkdir()) {
+				if (!tempConfigDir.mkdirs()) {
 					throw new MojoExecutionException("Unable to create temporary directory " + tempConfigDir.getAbsolutePath() + "\n");
 				}
 				moduleInstaller.installAndUnpackModule(configArtifact, tempConfigDir.getAbsolutePath());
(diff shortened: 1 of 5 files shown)


sdk-commons/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
```diff
@@ -127,15 +127,12 @@ public Artifact getParentArtifact() {
...
         if (StringUtils.isBlank(parentArtifactId)) {
-            log.warn("parent.artifactId  missing");
             missingCount++;
         }
         if (StringUtils.isBlank(parentGroupId)) {
-            log.warn("parent.groupId is missing");
             missingCount++;
         }
         if (StringUtils.isBlank(parentVersion)) {
-            log.warn("parent.version is missing");
```
ibacher, line 138: Not sure why we're removing these lines?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -39,6 +37,7 @@
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.UUID;
```
ibacher, line 40: Looks like we can drop this import?

ibacher (approved):

wikumChamith (approved):

sdk-commons/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
mseaton: @ibacher / @wikumChamith ... Yeah, not related to this PR, but I removed them because they seem incorrect and I noticed them in my log and investigated.  If I have no parent properties at all in my distro, I get these 3 warning messages in my console, which is wrong.  We can certainly keep these warnings in place if we first check that at least one property is entered, but honestly I don't understand the logic, because the conditional statement below seems to render all this moot?  Based on the conditional below, if you have 1 or 2 missing, then things fail.  So the only allowed situation is 0 properties or all 3 properties.  So why are we logging a warning above exactly?

maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
mseaton: Done, thanks for noticing

mseaton:
After reviewing with @ibacher , I think we are going to table this in favor of deprecating this functionality and focusing on support for content packages.  I'm going to close this out, though retain the right to reopen.

PR #302, opened 2024-11-08
SDK-342 - SDK fails if distro.properties supplies type or group for c…

mseaton:
…ontent package

see https://openmrs.atlassian.net/browse/SDK-342

### DIFF
--- a/sdk-commons/src/main/java/org/openmrs/maven/plugins/model/BaseSdkProperties.java
+++ b/sdk-commons/src/main/java/org/openmrs/maven/plugins/model/BaseSdkProperties.java
@@ -154,9 +154,14 @@ public abstract class BaseSdkProperties {
 		List<Artifact> artifacts = new ArrayList<>();
 		for (Object keyObject : getAllKeys()) {
 			String key = keyObject.toString();
-			if (key.startsWith(TYPE_CONTENT + ".")) {
-				artifacts.add(new Artifact(checkIfOverwritten(key, ARTIFACT_ID), getParam(key),
-				        checkIfOverwritten(key, GROUP_ID), checkIfOverwritten(key, TYPE)));
+            String artifactType = getArtifactType(key);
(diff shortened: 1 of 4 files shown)


sdk-commons/src/main/java/org/openmrs/maven/plugins/model/BaseSdkProperties.java
```diff
@@ -154,9 +154,14 @@ public List<Artifact> getContentArtifacts() {
 		List<Artifact> artifacts = new ArrayList<>();
```
wikumChamith, line 154: @mseaton wouldn't it be better to consolidate duplicated code for extracting different artifact types into a single reusable method?

```suggestion
    public List<Artifact> getWarArtifacts(){
        return getArtifacts(TYPE_WAR);
    }

	public List<Artifact> getConfigArtifacts() {
        return getArtifacts(TYPE_CONFIG);
    }

	public List<Artifact> getContentArtifacts() {
        return getArtifacts(TYPE_CONTENT);
    }

    private List<Artifact> getArtifacts(String typeContent) {
        List<Artifact> artifacts = new ArrayList<>();
        for (Object keyObject : getAllKeys()) {
            String key = keyObject.toString();
String artifactType = getArtifactType(key);
            if (artifactType.equals(typeContent)) {
                artifacts.add(new Artifact(
checkIfOverwritten(key, ARTIFACT_ID),
getParam(key),
                        checkIfOverwritten(key, GROUP_ID),
checkIfOverwritten(key, TYPE)
));
            }
        }
        return artifacts;
    
```

mseaton: Indeed @wikumChamith there are many, many opportunities to work on tech debt in the SDK.  This initial PR is attempting to just fix the underlying issue.  More than happy to make broader refactoring improvements in other commits / tickets.

PR #303, opened 2024-11-08
SDK-344 - SDK should support including spa from a maven artifact

mseaton:
see https://openmrs.atlassian.net/browse/SDK-344

### DIFF
--- a/integration-tests/src/test/resources/integration-test/openmrs-distro-spa-artifacts.properties
+++ b/integration-tests/src/test/resources/integration-test/openmrs-distro-spa-artifacts.properties
@@ -8,6 +8,6 @@ omod.spa=2.0.0
 spa.artifactId=openmrs-frontend-zl
 spa.groupId=org.pih.openmrs
 spa.version=1.3.0
-spa.include=openmrs-frontend-zl-1.3.0
+spa.includes=openmrs-frontend-zl-1.3.0
 
 db.h2.supported=false
--- a/integration-tests/src/test/resources/integration-test/openmrs-distro-spa-artifacts.properties
+++ b/integration-tests/src/test/resources/integration-test/openmrs-distro-spa-artifacts.properties
@@ -1,0 +1,13 @@
+name=Spa Artifact Example
+version=1.0.0-SNAPSHOT
+
+war.openmrs=2.6.9
+
+omod.spa=2.0.0
+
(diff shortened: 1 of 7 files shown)


integration-tests/src/test/resources/integration-test/openmrs-distro-spa-artifacts.properties
```diff
@@ -0,0 +1,13 @@
...
+version=1.0.0-SNAPSHOT
+
+war.openmrs=2.6.9
+
+omod.spa=2.0.0
+
+spa.artifactId=openmrs-frontend-zl
+spa.groupId=org.pih.openmrs
+spa.version=1.3.0
+spa.include=openmrs-frontend-zl-1.3.0
```
mogoodrich, line 11: This is probably obvious and I'm just missing it, but why do we have to limit to "openmrs-frontend-zl-1.3.0"

mogoodrich (approved):

integration-tests/src/test/resources/integration-test/openmrs-distro-spa-artifacts.properties
mseaton: This is a hack/feature I had to add in, which I had originally hoped to avoid, because when we build our frontend zips we don't just include the files but we include them in a nested directory with this name.  So this is how I am providing an option to say "don't just unpack the entire zip as-is, but unpack it and copy in the contents of this subdirectory within the zip".

ibacher:
Not essential to this, but the unpack goal we’re using has an `includes` option which we could leverage. Same outcome, but involves less copying.

mseaton:
> Not essential to this, but the unpack goal we’re using has an `includes` option which we could leverage. Same outcome, but involves less copying.

Interesting - I'll look to see if I can switch this to use that, thanks @ibacher .  If nothing else, it gives me more confidence in my choice of property name, which I struggled over.   Maybe I'll update it from `include` to `includes`.

mseaton:
So, I tried out using the includes property of the unpack goal, and it did not work as I wanted.  As far as I can tell, there isn't a way to tell it to include the contents of a subdirectory but not the subdirectory itself.  So I kept my initial implementation.  I did change the property name from `include` to `includes` to leave room to support indicating multiple includes rather than just one, though the current implementation just assumes one.

PR #306, opened 2024-11-15
SDK-354 - Fixing bugs and adding test cases for the AddExclusions goal

mseaton:
see https://openmrs.atlassian.net/browse/SDK-354

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/AddExclusion.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/AddExclusion.java
@@ -15,6 +15,9 @@ import java.util.stream.Collectors;
 @Mojo(name = "exclusion", requiresProject = false)
 public class AddExclusion extends AbstractTask {
 
+    public static final String WARNING_PROPERTY_NOT_IN_PARENT = "The property is not included in the parent distro";
+    public static final String WARNING_NO_PARENT_DISTRO = "This distro properties file does not contain a valid parent distro";
+
     /**
      * Path to the openmrs-distro.properties file to modify
      */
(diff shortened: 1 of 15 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/AddExclusion.java
```diff
@@ -51,22 +54,23 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
             throw new MojoFailureException("Invalid distro file");
         }
 
-        Artifact distroArtifact = originalDistroProperties.getParentArtifact();
-        if (StringUtils.isNotBlank(distroArtifact.getArtifactId()) && StringUtils.isNotBlank(distroArtifact.getGroupId()) && StringUtils.isNotBlank(distroArtifact.getVersion())) {
-            DistroProperties distroProperties = distroHelper.resolveParentArtifact(distroArtifact, new File(distro).getParentFile(), originalDistroProperties, null);
```
mseaton, line 56: In my testing, this was actually not getting the full set of parent properties.  This code also does not resolve full ancestors (eg. if the parent has a parent).  The change below intends to fix this.

maven-plugin/src/main/java/org/openmrs/maven/plugins/AddExclusion.java
```diff
@@ -51,22 +54,23 @@ public void executeTask() throws MojoExecutionException, MojoFailureException {
...
             if (StringUtils.isBlank(property)) {
-                List<String> currentExclusions = distroProperties.getExclusions();
-                List<String> options = distroProperties.getPropertyNames().stream().filter(prop -> !currentExclusions.contains(prop)).collect(Collectors.toList());
-                property = wizard.promptForMissingValueWithOptions("Enter the property you want to exclude",
-                        null, null, options);
-            } else {
-                if (!distroProperties.getPropertyNames().contains(property)) {
-                    wizard.showWarning("The property is not included in the parent distro");
+                List<String> currentExclusions = originalDistroProperties.getExclusions();
+                List<String> options = parentProperties.getPropertyNames().stream().filter(prop -> !currentExclusions.contains(prop)).collect(Collectors.toList());
```
mseaton, line 63: Notice the difference here from above - the _options_ need to come from the parent properties, but the current _exclusions_ need to come from the child properties

sdk-commons/src/main/java/org/openmrs/maven/plugins/model/DistroProperties.java
```diff
@@ -109,46 +109,59 @@ private String extractPropertyName(String key) {
         }
     }
```
mseaton, line 111: We can review this in more detail in SDK-348, but for now I have consolidated the two methods of specifying a parent distro so that both are equally supported.  See change below.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -165,39 +164,61 @@ public void savePropertiesToServer(DistroProperties properties, Server server) t
...
 		String artifactId = split[split.length - 2];
 		String version = split[split.length - 1];
+		return normalizeArtifact(new Artifact(artifactId, version, groupId), versionsHelper);
+	}
 
-		if (versionsHelper != null && version.contains(SDKConstants.LATEST_VERSION_BATCH_KEYWORD)) {
-			if (version.equalsIgnoreCase(SDKConstants.LATEST_SNAPSHOT_BATCH_KEYWORD)) {
-				version = versionsHelper.getLatestSnapshotVersion(new Artifact(artifactId, version, groupId));
-			} else if (version.equalsIgnoreCase((SDKConstants.LATEST_VERSION_BATCH_KEYWORD))) {
-				version = versionsHelper.getLatestReleasedVersion(new Artifact(artifactId, version, groupId));
```
mseaton, line 181: There is a lot of scattered code in the SDK that does "special purpose stuff".  One example of that is resolving "referenceapplication" to either "referenceapplication-package" or "referenceapplication-distro" depending on the version.  This was not being applied everywhere equally, so this is an effort to create a method for doing this that can be called from various places to gain consistency.  More refactoring and consolidation down the road will help, but this is a start.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -527,12 +548,15 @@ public Properties getFrontendProperties(Server server) throws MojoExecutionExcep
 	}
 
 	public Properties getFrontendPropertiesForServer(Artifact artifact, File directory) throws MojoExecutionException {
-		if (new Version(artifact.getVersion()).higher(new Version("3.0.0-beta.16"))) {
-			return getFrontendProperties(artifact, directory);
-		} else {
-			return PropertiesUtils.getFrontendPropertiesFromSpaConfigUrl(
-					"https://raw.githubusercontent.com/openmrs/openmrs-distro-referenceapplication/" + artifact.getVersion() + "/frontend/spa-build-config.json");
+		if (artifact.getArtifactId().equals("referenceapplication-distro")) {
```
mseaton, line 551: This doesn't work for all cases.  Specifically, I was testing with a refapp 2.x distro as the parent, and in trying to do that it was failing here, as no resource in github exists for spa-build-config.json in a 2.x release/tag of the refapp, as far as I can tell.

So I wrapped this in a check to only resolve these properties if we are dealing with the "referenceapplication-distro" artifact, which is what is used starting in refapp 3.x.  Can you confirm this is what we want @ibacher  and @dkayiwa ?  This whole downloading frontend config from github is...odd.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -555,6 +579,38 @@ public Properties getArtifactProperties(Artifact artifact, File directory, Strin
 		return properties;
 	}
 
+	// TODO: This needs unit tests, but to do so will require more refactoring to allow downloadDistroProperties to be mocked.
```
mseaton, line 582: Per my comment, this needs to have proper tests around it, but I'm punting that to a later date due to the potential complexity of doing this in a unit test and needing to mock methods within the class being tested.

PR #308, opened 2024-11-18
SDK-358 - Add integration test framework for helper classes

mseaton:
see https://openmrs.atlassian.net/browse/SDK-358


maven-plugin/pom.xml
```diff
@@ -226,7 +232,35 @@
...
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>3.4.2</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
mseaton, line 263: This adds all of the test classes to their own library, so this library can be added to the classpath of the pom used by the Maven verifier during the integration tests, which runs outside of this project

maven-plugin/src/main/java/org/openmrs/maven/plugins/AbstractTask.java
```diff
@@ -166,6 +176,16 @@ public AbstractTask(AbstractTask other) {
...
+		if (mavenEnvironment == null) {
+			mavenEnvironment = new MavenEnvironment();
+			mavenEnvironment.setMavenProject(mavenProject);
+			mavenEnvironment.setMavenSession(mavenSession);
+			mavenEnvironment.setSettings(settings);
+			mavenEnvironment.setArtifactMetadataSource(artifactMetadataSource);
+			mavenEnvironment.setArtifactFactory(artifactFactory);
+			mavenEnvironment.setPluginManager(pluginManager);
+			mavenEnvironment.setWizard(wizard);
+		}
```
mseaton, line 188: See the new `MavenEnvironment` class below - basically this just allows us to encapsulate everything provided by the Maven process during plugin execution in a single object that can be passed around and depended upon.

maven-plugin/src/test/java/org/openmrs/maven/plugins/AbstractMavenIT.java
```diff
@@ -0,0 +1,55 @@
...
+			addTaskParam("className", className);
+			addTaskParam("methodName", testMethod);
+			addTaskParam(BATCH_ANSWERS, getAnswers());
+			addTaskParam("testMode", "true");
+			String plugin = resolveSdkArtifact();
+			verifier.executeGoal(plugin + ":" + InvokeMethod.NAME);
+		}
+		else {
+			testFunction.executeTest();
+		}
```
mseaton, line 42: So the general idea here, is that a particular integration test will wrap it's test functionality in this `executeTest` method.  When this is run, the mavenEnvironment will be null, and this will trigger the new "InvokeMethod" plugin to run it, by creating a new instance of this same test class, and executing this same method, but this time with the mavenEnvironment populated from the InvokeMethod plugin.

maven-plugin/src/test/java/org/openmrs/maven/plugins/utility/ArtifactHelperIT.java
```diff
@@ -0,0 +1,48 @@
...
+	@Test
+	public void test_downloadModuleWithDefaultName() throws Exception {
+		executeTest(() -> {
+			ArtifactHelper artifactHelper = new ArtifactHelper(getMavenEnvironment());
+			Artifact artifact = new Artifact("idgen-omod", "4.14.0", "org.openmrs.module", "jar");
+			artifactHelper.downloadArtifact(artifact, getMavenTestDirectory(), null);
+			File expectedFile = new File(getMavenTestDirectory(), "idgen-omod-4.14.0.jar");
+			assertTrue(expectedFile.exists());
+		});
+	}
```
mseaton, line 27: And so this is an example of how we would run such a test.  By wrapping the code in the `executeTest` block, it will cause the code in that block to be executed by a bare-bones plugin that exists only to execute the provided block.

maven-plugin/src/test/resources/integration-test/invokeIT/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+    <artifactId>openmrs-sdk-integration-test-project</artifactId>
+    <version>1.0</version>
+    <packaging>pom</packaging>
+
+    <!--
+        This makes the test classes defined within this project available on the Plugin classpath when it is executed
+        Any new test libraries that are referenced by tests that are use this pom (eg. those that extend AbstractMavenIT)
+        need to be added as dependencies must be added here so the plugin can load and execute the tests successfully
+        See:  https://maven.apache.org/guides/mini/guide-maven-classloading.html
+    -->
```
mseaton, line 18: This took me the longest time to figure out, but finally was able to do so.  See comment above.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.twdata.maven.mojoexecutor.MojoExecutor.goal;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.groupId;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.name;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.plugin;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.version;
+
+/**
+ * The purpose of this class is to handle all interactions with Maven that require retrieving artifacts from the Maven repository
+ */
+public class ArtifactHelper {
```
mseaton, line 28: One of my goals that led me to doing this in the first place was to start identifying functionality that we might refactor into common libraries.  This is the start of that effort, and the integration test that corresponds with it is necessary to make sure it is operating as expected.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/MavenEnvironment.java
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.maven.artifact.metadata.ArtifactMetadataSource;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.plugin.BuildPluginManager;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.settings.Settings;
+
+/**
+ * Component that allows access to the Maven components set within the current execution environment
+ */
+@Data
```
mseaton, line 14: I love lombok, we should use it everywhere.

maven-plugin/src/main/java/org/openmrs/maven/plugins/InvokeMethod.java
```diff
@@ -0,0 +1,93 @@
...
+
+        if (setterMethod != null) {
+            try {
+                setterMethod.invoke(instance, getMavenEnvironment());
+                wizard.showMessage("Instance populated with maven environment");
+            }
+            catch (Exception e) {
+                    throw new MojoExecutionException("Error executing method: " + methodName, e);
+            }
+        }
```
mseaton, line 74: This code is admittedly ugly.  I probably should have just found a library to do this all more elegantly.  In any event, this plugin will instantiate the given class, set the `mavenEnvironment` property on it if available, and the call the given method.  This is what allows us to test functionality that requires a Maven Environment outside of individual plugins that consume that functionality (aside from this one, which is necessary).

PR #314, opened 2024-11-26
SDK-343 - Setup task presents both the release and latest snapshot of…

mseaton:
… the same version as options

see https://openmrs.atlassian.net/browse/SDK-343

### DIFF
--- a/sdk-commons/src/test/java/org/openmrs/maven/plugins/utility/VersionsHelperTest.java
+++ b/sdk-commons/src/test/java/org/openmrs/maven/plugins/utility/VersionsHelperTest.java
@@ -18,63 +18,70 @@ public class VersionsHelperTest {
     VersionsHelper helper;
 
     @Before
-    public void before(){
+    public void before() {
         helper = new VersionsHelper(null, null, null, null);
     }
 
-    private List<ArtifactVersion> createTestVersions(String... versions){
+    private List<ArtifactVersion> createTestVersions(String... versions) {
         List<ArtifactVersion> list = new ArrayList<>();
-        for(String version: versions) {
+        for (String version : versions) {
             list.add(new DefaultArtifactVersion(version));
         }
         return list;
     }
 
     @Test
-    public void getLatestReleasedVersion_shouldNotIncludeSnapshots(){
+    public void getLatestReleasedVersion_shouldNotIncludeSnapshots() {
         String res = helper.getLatestReleasedVersion(
                 createTestVersions("1.8.8", "1.8.3", "1.8.10-SNAPSHOT", "1.8.9-alpha"));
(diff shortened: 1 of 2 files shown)


sdk-commons/src/test/java/org/openmrs/maven/plugins/utility/VersionsHelperTest.java
```diff
@@ -18,63 +18,70 @@ public class VersionsHelperTest {
...
         assertThat(res, hasItem("1.8.10-SNAPSHOT"));
     }
+
     @Test
-    public void getSuggestedVersions_containGivenNumberOfReleases(){
+    public void getSuggestedVersions_containGivenNumberOfReleases() {
         List<String> res = helper.getSuggestedVersions(
                 createTestVersions("1.8.8", "1.8.3", "1.8.10-SNAPSHOT", "1.8.9-alpha"), 1);
-        assertThat(res, hasItem("1.8.9-alpha"));
+        assertThat(res, hasItem("1.8.10-SNAPSHOT"));
```
mseaton, line 58: This seems like a bug in the test, which allowed bugs in the code to pass.  If asking for suggested versions with a max number of 1, it should return the latest suggested version, and 1.8.10 > 1.8.9

sdk-commons/src/test/java/org/openmrs/maven/plugins/utility/VersionsHelperTest.java
```diff
@@ -18,63 +18,70 @@ public class VersionsHelperTest {
...
     }
+
     @Test
-    public void getSuggestedVersions_shouldReturnSortedList(){
+    public void getSuggestedVersions_shouldReturnSortedList() {
         List<String> res = helper.getSuggestedVersions(
-                createTestVersions("1.1.2", "1.5.6", "1.5.7-SNAPSHOT" , "1.1.2-alpha", "1.1.0-SNAPSHOT", "1.3.0-SNAPSHOT", "1.3.0", "1.1.0"), 6);
-        assertThat(res, contains("1.5.7-SNAPSHOT", "1.5.6", "1.3.0", "1.3.0-SNAPSHOT", "1.1.2"));
+                createTestVersions("1.1.2", "1.5.6", "1.5.7-SNAPSHOT", "1.1.2-alpha", "1.1.0-SNAPSHOT", "1.3.0-SNAPSHOT", "1.3.0", "1.1.0"), 6);
+        assertThat(res, contains("1.5.7-SNAPSHOT", "1.5.6", "1.3.0", "1.1.2"));
```
mseaton, line 66: Ultimately, this bug in this test was actually the case this ticket was aiming to fix.  The suggested versions should _not_ include both a release and a snapshot with the same incremental version.  After fixing the code, this test broke, then I fixed the test.

PR #315, opened 2024-11-26
SDK-361 and SDK-356 - Update wizard to support custom distributions and more platform and distribution versions

mseaton:
see https://openmrs.atlassian.net/browse/SDK-361
see https://openmrs.atlassian.net/browse/SDK-356

See ticket details for description of changes.

### DIFF
--- a/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
+++ b/maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
@@ -16,7 +16,6 @@ import org.openmrs.maven.plugins.model.Version;
 import org.openmrs.maven.plugins.utility.ContentHelper;
 import org.openmrs.maven.plugins.utility.DBConnector;
 import org.openmrs.maven.plugins.utility.DistributionBuilder;
-import org.openmrs.maven.plugins.utility.DistroHelper;
 import org.openmrs.maven.plugins.utility.SDKConstants;
 import org.openmrs.maven.plugins.utility.ServerHelper;
 
@@ -40,14 +39,8 @@ import java.util.List;
 
 import static org.openmrs.maven.plugins.model.Artifact.GROUP_DISTRO;
 import static org.openmrs.maven.plugins.utility.SDKConstants.PLATFORM_ARTIFACT_ID;
-import static org.openmrs.maven.plugins.utility.SDKConstants.REFAPP_2X_ARTIFACT_ID;
-import static org.openmrs.maven.plugins.utility.SDKConstants.REFAPP_2X_GROUP_ID;
 import static org.openmrs.maven.plugins.utility.SDKConstants.REFAPP_2X_PROMPT;
-import static org.openmrs.maven.plugins.utility.SDKConstants.REFAPP_2X_TYPE;
-import static org.openmrs.maven.plugins.utility.SDKConstants.REFAPP_3X_ARTIFACT_ID;
-import static org.openmrs.maven.plugins.utility.SDKConstants.REFAPP_3X_GROUP_ID;
 import static org.openmrs.maven.plugins.utility.SDKConstants.REFAPP_3X_PROMPT;
-import static org.openmrs.maven.plugins.utility.SDKConstants.REFAPP_3X_TYPE;
 import static org.openmrs.maven.plugins.utility.SDKConstants.SETUP_DEFAULT_PLATFORM_VERSION;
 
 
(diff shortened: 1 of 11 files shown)


maven-plugin/src/main/java/org/openmrs/maven/plugins/Setup.java
```diff
@@ -205,53 +196,65 @@ private DistroProperties resolveDistroProperties(Server server) throws MojoExecu
...
 
-		options.add(O3_DISTRIBUTION);
-		options.add(O2_DISTRIBUTION);
+		options.add(REFAPP_2X_PROMPT);
+		options.add(REFAPP_3X_PROMPT);
 		options.add(PLATFORM);
-		String choice = wizard.promptForMissingValueWithOptions(SETUP_SERVERS_PROMPT, null, null, options);
 
+		String customDistroMessage = "Please specify distribution artifact";
+		String customDistroDefault = SDKConstants.REFERENCEAPPLICATION_2_4;
```
mseaton, line 220: By adding these to this top-level wizard step, we move the option to specify a custom distribution out from under the RefApp 2.x section and into the top-level section where it belongs.  This then allows us to focus the refapp 2.x section on paging through the published versions in Maven

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -233,6 +218,49 @@ public String promptForMissingValueWithOptions(String message, String value, Str
...
+
+		if (HIGHER_VERSIONS.equals(answer)) {
+			return promptForArtifactVersion(message, options, pageNum - 1, pageSize, otherMessage);
+		}
+		else if (LOWER_VERSIONS.equals(answer)) {
+			return promptForArtifactVersion(message, options, pageNum + 1, pageSize, otherMessage);
+		}
+
+		return answer;
+	}
```
mseaton, line 262: This is the main code that now drives the Wizard when one is selecting a platform, 2.x refapp or 3.x refapp version.  This downloads all suggested versions and allows paging through all of them, rather than just presenting the most recent 5 or 6 or whatever.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -413,15 +434,32 @@ public String promptForExistingServerIdIfMissing(String serverId) throws MojoExe
...
 
 	@Override
-	public String promptForPlatformVersion(List<String> versions) throws MojoExecutionException {
-		return promptForMissingValueWithOptions(PLATFORM_VERSION_PROMPT,
-				null, "version", versions, "Please specify platform version", null);
+	public Artifact promptForRefApp3xArtifact(VersionsHelper versionsHelper) throws MojoExecutionException {
+		Artifact artifact = new Artifact(REFAPP_3X_ARTIFACT_ID, "3.0.0", REFAPP_3X_GROUP_ID, REFAPP_3X_TYPE);
+		String version = promptForArtifactVersion("Please choose a " + REFAPP_3X_PROMPT + " version", artifact, null, versionsHelper);
+		artifact.setVersion(version);
+		return artifact;
```
mseaton, line 462: Much more straightforward and simplified methods here for prompting for versions, and able to remove the web of other methods below that are no longer used or needed.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -651,168 +689,6 @@ private void savePropertiesChangesToFile(Properties properties, File file)
...
-			return optionsMap;
-		}
-
-		{
-			Artifact artifact = new Artifact(SDKConstants.REFAPP_DISTRO, "3.0.0", "org.openmrs.distro", "zip");
-			for (ArtifactVersion version : versionsHelper.getAllVersions(artifact, MAX_OPTIONS_SIZE)) {
-				if (!version.toString().endsWith("-SNAPSHOT") && optionsMap.size() < MAX_OPTIONS_SIZE) {
-					optionsMap.put(String.format(optionTemplate, version),
-							artifact.getGroupId() + ":" + artifact.getArtifactId() + ":" + version);
-				}
```
mseaton, line 809: This is something that was easier NOT to support in this change, and it didn't seem like something we need anymore now that 3.0.0 has been released.  This particular distro artifact was only ever a snapshot/beta/alpha of 3.0.0, so there should not be any need for it anymore.  Someone could still choose to install it via the custom distribution option, by specifying the full maven coordinate string, but I don't think we need this sitting in the list of versions in the Wizard, so I removed it.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -1053,17 +929,6 @@ public void promptForDockerizedDb(Server server, DockerHelper dockerHelper, Stri
...
-	private String getDefaultUri(String dockerHost) {
-		if (SystemUtils.IS_OS_LINUX) {
-			return SDKConstants.URI_MYSQL;
-		} else {
-			int beginIndex = dockerHost.indexOf("//");
-			int endIndex = dockerHost.lastIndexOf(":");
-			String dockerMachineIp = dockerHost.substring(beginIndex + 2, endIndex);
-			return String.format(SDKConstants.URI_MYSQL_DOCKER, dockerMachineIp);
-		}
-	}
```
mseaton, line 1065: Unrelated, this method wasn't used, so I cleaned it out.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java
```diff
@@ -1084,20 +949,12 @@ public void promptForDbCredentialsIfMissing(Server server) throws MojoExecutionE
...
 		Path openMRS = Server.getServersPath();
 		Map<Long, String> sortedMap = new TreeMap<>(Collections.reverseOrder());
-		try (DirectoryStream<Path> subDirectories = Files.newDirectoryStream(openMRS, new DirectoryStream.Filter<Path>() {
-
-			@Override
-			public boolean accept(Path entry) {
-				return entry.toFile().isDirectory();
-			}
-		})) {
+		try (DirectoryStream<Path> subDirectories = Files.newDirectoryStream(openMRS, entry -> entry.toFile().isDirectory())) {
```
mseaton, line 957: IDE-suggested code cleanup.

PR #318, opened 2024-12-10
SDK-365 and SDK-369 - Ensure content package validation occurs prior …

mseaton:
…to destructive changes, and do not automatically change distribution/server based on content package requirements.

see https://openmrs.atlassian.net/browse/SDK-365
see https://openmrs.atlassian.net/browse/SDK-369


integration-tests/src/test/java/org/openmrs/maven/plugins/AbstractSdkIT.java
```diff
@@ -171,6 +171,22 @@ public String setupTestServer(String distro) throws Exception{
...
+            addAnswer("y");
+            addTaskParam(setupContentPackage, BATCH_ANSWERS, getAnswers());
+            setupContentPackage.executeGoal(resolveSdkArtifact() + ":build");
+        }
+        finally {
+            cleanAnswers();
+            verifier.resetStreams();
+        }
+    }
+
```
mseaton, line 189: This allows us to create any content packages we wish for testing, without relying on published versions in the cloud

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIT.java
```diff
@@ -527,6 +528,35 @@ public void setup_shouldInstallWithContentPackageWithNoNamespace() throws Except
...
+        }
+        assertNotNull(exception);
+        assertLogContains("requires module org.openmrs.module:fhir2-omod version >=1.8.0");
+        assertLogContains("requires module org.openmrs.module:webservices.rest-omod version >=2.42.0");
+        assertLogContains("requires module org.openmrs.module:openconceptlab-omod version >=2.3.0");
+        assertLogContains("requires module org.openmrs.module:initializer-omod version >=2.5.2");
+        assertLogContains("requires module org.openmrs.module:o3forms-omod version >=2.2.0");
+        assertLogContains("requires esm @openmrs/esm-patient-chart-app version >=8.1.0");
+    }
+
```
mseaton, line 559: I have added integration tests for the 3 impacted goals with respect to content package validation.  In each case, the new default behavior is to fail, and provide a clear message as to why the build failed and what is needed to fix it.

maven-plugin/src/main/java/org/openmrs/maven/plugins/ServerUpgrader.java
```diff
@@ -54,6 +54,9 @@ public void upgradeToDistro(Server server, Distribution distribution, boolean ig
 				return;
 			}
 		}
+
+		parentTask.distroHelper.validateDistribution(distroProperties);
```
mseaton, line 58: Per SDK-365, this moves the validation earlier in the process.  Previously this was not running until after war/module/etc changes had already been applied to the server.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -299,213 +303,152 @@ public void saveDistroPropertiesTo(File destination, String distro) throws MojoE
...
-			downloadContentPackages(tempDirectory, distroProperties);
-
-		} catch (IOException e) {
-			throw new MojoExecutionException("Failed to process content packages", e);
-		} finally {
-			if (tempDirectory != null && tempDirectory.exists()) {
-				try {
-					FileUtils.deleteDirectory(tempDirectory);
-				} catch (IOException e) {
-					log.warn("Failed to delete temporary directory: {}", tempDirectory.getAbsolutePath(), e);
```
mseaton, line 332: This is a convenience method that gets the missing dependencies, iterates over them, formats these into friendly error messages for display, and throws an exception with this.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -299,213 +303,152 @@ public void saveDistroPropertiesTo(File destination, String distro) throws MojoE
...
-	 *                                such as missing or invalid {@code content.properties} files, or any IO issues.
+	 * This inspects the distribution based on distro properties, and indicates any incompatibilities in the declared version
+	 * Currently, this only reviews dependencies declared within content packages, though this could be expanded to validate
+	 * modules based on config.xml, etc.
 	 */
-	public void downloadContentPackages(File contentPackageZipFile, DistroProperties distroProperties)
-			throws MojoExecutionException {
-
-		Properties contentProperties = new Properties();
-
```
mseaton, line 363: This is what actually retrieves and calculates missing dependencies.  If we wanted to add back in some interactivity with the wizard and allow the user to choose whether to upgrade their server/distribution to the necessary artifacts and versions, we would change the goal from calling the `validateDistribution` method to calling this method and then using the list of missing dependencies to drive the wizard and support the user upgrading.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -299,213 +303,152 @@ public void saveDistroPropertiesTo(File destination, String distro) throws MojoE
...
-			if (dependency.startsWith("omod.") || dependency.startsWith("owa.") || dependency.startsWith("war")
-					|| dependency.startsWith("spa.frontendModule") || dependency.startsWith("content.")) {
-				String versionRange = contentProperties.getProperty(dependency);
-				String distroVersion = distroProperties.getParam(dependency);
-
-				if (distroVersion == null) {
-					String latestVersion = findLatestMatchingVersion(dependency, versionRange);
-					if (latestVersion == null) {
-						throw new MojoExecutionException(
-								"No matching version found for dependency " + dependency + " in " + zipFileName);
```
mseaton, line 453: This method is a bit of a hack to support what PIH is currently doing, which is building our frontend in a separate project and publishing it to Maven, and then including that Maven zip in our distribution.  In this case, we don't have the spa-assemble or spa-build config files in that artifact (though as an alternative to this below, we could potentially add it in).  This downloads the artifact, unpacks it, and inspects the file/directory contents, translating the directory names back into ESM names and versions.  Interested in review/thoughts on this in particular @ibacher .  There is an integration test in `DistroHelperIT` that demonstrates this.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DistroHelper.java
```diff
@@ -299,213 +303,152 @@ public void saveDistroPropertiesTo(File destination, String distro) throws MojoE
...
-		try {
-			boolean inRange = semverVersion.satisfies(contentDependencyVersionRange.trim());
-			if (!inRange) {
-				throw new MojoExecutionException("Incompatible version for " + contentDependencyKey + " in content package "
-						+ contentPackageName + ". Specified range: " + contentDependencyVersionRange
-						+ ", found in distribution: " + distroPropertyVersion);
-			}
-		} catch (SemverException e) {
-			throw new MojoExecutionException("Invalid version range format for " + contentDependencyKey
-					+ " in content package " + contentPackageName + ": " + contentDependencyVersionRange, e);
```
mseaton, line 487: Per slack discussion this new method is what checks for version incompatibility, accounting for things like "next", and also allowing snapshot and pre-releases that are otherwise higher than a release version to satisfy that release version.  See unit test in `DistroHelperTest` for examples.

PR #357, opened 2026-01-06
[SDK-396] SDK Setup fails on Linux due to Docker/IPv6 parsing error

MonotonousHarsh:
Fixed the SDK setup process crashing on Linux environments. 
This PR addresses two setup blockers I encountered sequentially on Linux:
1. Changed Docker bindings from `localhost` to `127.0.0.1` to prevent IPv6 `ParseAddr` errors.
2. Added a connection retry mechanism because the MySQL container initialization caused a `Socket.connect` timeout before the DB was fully ready.

## Issue I worked on
see https://issues.openmrs.org/browse/SDK-396

- [x] My code follows the code style of this project.
- [x] I have read the contributing guidelines.
- [x] I have added/updated necessary tests.
- [x] All new and existing tests passed.

### DIFF
--- a/docker-maven-plugin/src/main/java/org/openmrs/maven/plugins/AbstractDockerMojo.java
+++ b/docker-maven-plugin/src/main/java/org/openmrs/maven/plugins/AbstractDockerMojo.java
@@ -26,7 +26,7 @@ abstract class AbstractDockerMojo extends AbstractMojo {
     protected static final String DEFAULT_MYSQL_EXPOSED_PORT = "3308";
     protected static final String DEFAULT_MYSQL_PASSWORD = "Admin123";
     protected static final String MYSQL_8_4_1 = "mysql:8.4.1";
-    protected static final String DEFAULT_MYSQL_DB_URI = "jdbc:mysql://localhost:" + DEFAULT_MYSQL_EXPOSED_PORT + "/";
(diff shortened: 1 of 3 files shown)


MonotonousHarsh:
@jwnasambu Yes, I have updated the PR title to include the Jira Ticket ID (SDK-396). The CI checks are passing. Is there anything else needed before this can be reviewed?

ibacher:
Please do not delete the PR template. It's there to make reviewer's lives easier. Without it, it's harder to understand the proposed changes. Similarly, this PR batches a number of changes for various things together into one go, which is discouraged.

ibacher:
> I grouped them because I encountered both crashes sequentially while trying to get the SDK to run on my local Linux environment 

Even so, it's easier to evaluate them if they're separated out. For example 15 retries with a 1 second pause between them feels like a seriously degrading user experience and needs a better approach, but I can't merge in the 127.0.0.1 thing because it's tied to that. But more than that, even if they happen in the same flow, they are two issues with two different root causes and, most likely two different fixes.

MonotonousHarsh:
@ibacher That makes complete sense, thank you for the explanation. You are right that they are distinct issues. Since this branch also has merge conflicts now, I am going to close this PR. I will open a fresh, clean PR containing only the 127.0.0.1 binding fix so we can get that resolved first. I will rethink the DB timeout issue and create a separate ticket for it later.

PR #371, opened 2026-05-27
SDK-400 - SDK should support more configurable Dockerfiles

mseaton:
See ticket comments

## Issue I worked on
see https://openmrs.atlassian.net/browse/SDK-400

- [x] I have **added tests** to cover my changes. (If you refactored


maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -371,20 +410,53 @@ private void copyDockerfile(File targetDirectory, DistroProperties distroPropert
...
-					copyBuildDistroResource("Dockerfile-jre11-bundled", new File(targetDirectory, "Dockerfile"));
-				} else {
-					copyBuildDistroResource("Dockerfile-jre11", new File(targetDirectory, "Dockerfile"));
+			String namespace = distroProperties.getParam(DOCKER_IMAGE_NAMESPACE, "openmrs");
+			String repository = distroProperties.getParam(DOCKER_IMAGE_REPOSITORY, "openmrs-core");
+			String dockerImageTag = distroProperties.getParam(DOCKER_IMAGE_TAG);
+			if (StringUtils.isBlank(dockerImageTag)) {
+				String openmrsVersion = distroProperties.getParam(DOCKER_IMAGE_OPENMRS_VERSION, null);
+				if (StringUtils.isNotBlank(openmrsVersion)) {
+					// NOTE: This will always treat a SNAPSHOT version as the latest snapshot in the minor release line, never an older snapshot
```
mseaton: Note this comment.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -371,20 +410,53 @@ private void copyDockerfile(File targetDirectory, DistroProperties distroPropert
...
 				}
 			}
-			else {
-				if (bundled) {
-					copyBuildDistroResource("Dockerfile-jre8-bundled", new File(targetDirectory, "Dockerfile"));
-				} else {
-					copyBuildDistroResource("Dockerfile-jre8", new File(targetDirectory, "Dockerfile"));
+			if (StringUtils.isBlank(dockerImageTag)) {
+				if (isPlatform2point5AndAbove(platformVersion)) {
+					dockerImageTag = "nightly-amazoncorretto-11";
```
mseaton: Should we be going to something higher for OpenMRS 3.0.0+?  What about 2.8 and 2.9 lines?

maven-plugin/src/main/resources/build-distro/web/Dockerfile-tomcat8
```diff
@@ -1,25 +0,0 @@
-FROM tomcat:8-jre8
```
mseaton, line 1: This file and the corresponding bundled one didn't seem to be referenced by anything, so I am deleting them without any replacement.

sdk-commons/src/test/java/org/openmrs/maven/plugins/model/DistroPropertiesTest.java
```diff
@@ -96,6 +97,50 @@ public void shouldGetContentPropertiesWithOverrides() {
...
+    }
+
+    @Test
+    public void resolveMavenPlaceholders_shouldLeaveDistroInternalRefsIntactForLaterResolution() throws MojoExecutionException {
+        // Simulates a distro.properties file that uses ${war.openmrs} (a distro property from the parent)
+        // alongside Maven properties.  Maven resolution must leave the distro ref intact so that
+        // resolveInternalPlaceholders can resolve it after the parent is merged.
+        Properties properties = new Properties();
+        properties.setProperty("name", "${project.name}");
+        properties.setProperty("docker.image.openmrsVersion", "${war.openmrs}");
```
mseaton: Main question is if this is what we'd want to do - support the ability for any property defined in the distro properties (or parent distro properties) to be available for variable interpolation alongside (and potentially clashing with) any maven defined variable.

Or would it be better to simplify to this very specific use case by handling something like `docker.image.openmrsVersion=war`, and if `war` is found, just using that the distro platform version?

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
ibacher: For 3, we definitely need to (3 has Java 21 as a minimum version due to Spring and Hibernate minimums). For 2.8 and 2.9, it's fine but would probably be best to run them on Java 21 as that has some performance improvements.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -73,6 +73,47 @@ public class BuildDistro extends AbstractTask {
...
+	 * The Docker image namespace (i.e. the registry organisation) to use in the generated Dockerfile.
+	 * Defaults to {@code openmrs}.
+	 */
+	static final String DOCKER_IMAGE_NAMESPACE = "docker.image.namespace";
+
+	/**
+	 * The Docker image repository to use in the generated Dockerfile.
+	 * Defaults to {@code openmrs-core}.
+	 */
+	static final String DOCKER_IMAGE_REPOSITORY = "docker.image.repository";
```
ibacher, line 93: Shouldn't this be something like `IMAGE_NAME`?

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -110,7 +151,7 @@ public class BuildDistro extends AbstractTask {
 	 * the war file in `/webapp/src/main/webapp/WEB-INF/bundledModules`
 	 */
 	@Parameter(defaultValue = "false", property = "bundled")
-	private boolean bundled;
+	boolean bundled;
```
ibacher, line 155: Is this scope change needed?

ibacher (approved):
Code mostly LGTM. Two small things.

mogoodrich (commented):
Not too familiar with this, so will generally defer to Daniel and Ian, but conceptually these changes seem great.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
mseaton: Yeah, I wasn't sure what the different components of the image name were called, so I went by this:  https://docs.docker.com/reference/cli/docker/image/tag/ .  I can change it to `docker.image.name` if you think that's better, but I feel like the entire contatenated thing is the name

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
mseaton: I can change it back and see.  This leaked in from AI when I asked it to generate tests, so I figured it was needed for that, but I'll check

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
mseaton: I can certainly change for 3, since it just won't work.  I can also change for lower versions though that would be a change in behavior.

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
ibacher: Ah, fair... As long as we're following some kind of Docker guidance that sounds better to me!

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
mseaton: OK, so I think things are already worse than I had thought.  We don't actually have (AFAICT) any Docker images tagged as `openmrs/openmrs-core:nightly-amazoncorretto-11` or `openmrs/openmrs-core:nightly-amazoncorretto-8`, so do the Docker images that are generated by the SDK even work at all?  I have an alternative approach committed here that I am testing out

ibacher: Yeah... I think we might've fallen victim to Docker Hub's aggressive pruning. I also don't see any images for 2.6 - 2.8 with nightly tags...

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -319,9 +364,9 @@ private String buildDistro(File targetDirectory, Distribution distribution) thro
 
 		wizard.showMessage("Creating Docker Compose configuration...\n");
 		String distroVersion = adjustImageName(distroProperties.getVersion());
-		writeDockerCompose(targetDirectory, distroVersion);
-		writeReadme(targetDirectory, distroVersion);
-		if(!isAbovePlatform2point0) {
+		writeDockerCompose(targetDirectory, distroVersion, platformVersion);
+		writeReadme(targetDirectory, distroVersion, platformVersion);
+		if(!isPlatform2point5AndAbove(platformVersion)) {
```
mseaton: This is a change from previous behavior.  Previously, we switched on 2.0+.  Now this switches on 2.5+.  So any distro with an OpenMRS war version < 2.5 will use one of the included Dockerfiles in the resources - either one for Tomcat 7 with Java 7 (1.x) or one for Tomcat 8 with Java 8 (2.x prior to 2.5).

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -361,43 +406,70 @@ private boolean isDockerComposeCreated(File targetDir) {
...
+			copyBuildDistroResource(dockerFile, new File(targetDirectory, "Dockerfile"));
+		}
+		else {
+			namespace = StringUtils.defaultIfBlank(namespace, "openmrs");
+			repository = StringUtils.defaultIfBlank(repository, "openmrs-core");
+			String dockerImageTag = distroProperties.getParam(DOCKER_IMAGE_TAG);
+			if (StringUtils.isBlank(dockerImageTag)) {
+				String defaultOpenmrsVersion = majorVersion + "." + minorVersion + ".x";
+				if (majorVersion == 2 && minorVersion == 5) {
+					defaultOpenmrsVersion += "-nightly";
```
mseaton: At least as of now, the 2.5.x tag in dockerhub has this -nightly appended to it.  It would be nice not to have this specific conditional here, especially because artifacts can change in Dockerhub.  Thoughts about this @ibacher and about the way this uses the <major>.<minor>.x tags in Dockerhub by default?  Is that something we can rely upon going forward?

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -361,43 +406,80 @@ private boolean isDockerComposeCreated(File targetDir) {
...
+			lines.add("COPY openmrs_core/openmrs.war /openmrs/distribution/openmrs_core/");
+			lines.add("COPY openmrs-distro.properties /openmrs/distribution/");
+			if (!bundled) {
+				lines.add("COPY openmrs_modules /openmrs/distribution/openmrs_modules");
+				lines.add("COPY openmrs_owas /openmrs/distribution/openmrs_owas");
+				lines.add("COPY openmrs_config /openmrs/distribution/openmrs_config");
+				lines.add("COPY openmrs_spa /openmrs/distribution/openmrs_spa");
+			}
+			lines.add("");
+			for (String extraProperty : distroProperties.getPropertiesNames()) {
```
mseaton, line 479: My goal here was to see if I could use the build-distro goal to build one of our pihemr distributions, and then simply use the docker and docker compose artifacts produced to successfully run this distribution.  Adding this in was necessary to accomplish this, since our pihemr distribution requires certain runtime properties to be in place to install correctly.  And since they are defined at the distro-properties-file-level, it seems appropriate to bake them into the image in the same way that the artifacts are, it can always be overridden

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -361,43 +406,80 @@ private boolean isDockerComposeCreated(File targetDir) {
...
 		}
 	}
 
 	private boolean isPlatform2point5AndAbove(Version platformVersion) {
-		 return isAbovePlatformVersion(platformVersion, 2, 5);
+		return isAtOrAbovePlatformVersion(platformVersion, 2, 5);
 	}
 
-	private boolean isAbovePlatformVersion(Version platformVersion, int majorVersion, int minorVersion) {
+	private boolean isAtOrAbovePlatformVersion(Version platformVersion, int majorVersion, int minorVersion) {
```
mseaton, line 598: This method name was misleading/wrong, so I changed it

maven-plugin/src/main/java/org/openmrs/maven/plugins/BuildDistro.java
```diff
@@ -433,7 +515,8 @@ private void writeTemplatedFile(File targetDirectory, String version, String pat
 		if (!compose.exists()) {
 			try (InputStream inputStream = composeUrl.openStream(); FileWriter composeWriter = new FileWriter(compose)) {
 				String content = IOUtils.toString(inputStream, StandardCharsets.UTF_8);
-				content = content.replaceAll("\\$\\{TAG:-nightly}", version);
+				String dbImage = isPlatform2point5AndAbove(platformVersion) ? "mariadb:10.11.7" : "mysql:5.6";
```
mseaton: When I tested this on earlier OpenMRS versions, they failed to start up using the mariadb image.  Switching to mysql:5.6 image works for me.  It's possible we could switch to mariadb earlier than 2.5, but not sure how important that is?

maven-plugin/src/main/resources/build-distro/web/Dockerfile-jre7
```diff
@@ -5,6 +5,8 @@ FROM tomcat:7-jre7
 COPY openmrs.war  /usr/local/tomcat/webapps/openmrs.war
 COPY modules /usr/local/tomcat/.OpenMRS/modules
 COPY owa /usr/local/tomcat/.OpenMRS/owa
+COPY frontend /usr/local/tomcat/.OpenMRS/frontend
+COPY configuration /usr/local/tomcat/.OpenMRS/configuration
```
mseaton, line 9: I added these just so we could have consistency across distributions

maven-plugin/src/main/resources/build-distro/web/Dockerfile-jre8
```diff
@@ -1,10 +1,25 @@
...
+COPY startup.sh /usr/local/tomcat/startup.sh
+
+RUN sed -i '/Connector port="8080"/a URIEncoding="UTF-8" relaxedPathChars="[]|" relaxedQueryChars="[]|{}^&#x5c;&#x60;&quot;&lt;&gt;"' /usr/local/tomcat/conf/server.xml
+
+RUN chmod +x /usr/local/tomcat/wait-for-it.sh
+RUN chmod +x /usr/local/tomcat/startup.sh
+
+# For documentation purpose only
+COPY openmrs-distro.properties /root/openmrs-distro.properties
+COPY Dockerfile /root/Dockerfile
```
mseaton, line 24: This is the same as the Dockerfile-tomcat8, just renamed to jre8 for consistency with the jre7 version

mseaton:
Closing this PR to reflect the fact that I have made a lot of changes since it was first opened.  Will open a new PR when I am ready

PR #373, opened 2026-06-12
Verify GPG signatures on downloaded artifacts during setup and build-distro

wikumChamith:

Verifies the GPG signature of each downloaded module/WAR/OWA against a bundled, fingerprint-pinned OpenMRS signing key during setup and build-distro. Unsigned releases warn and continue; an invalid signature fails the build. On by default, disable with `-DverifySignatures=false`.

## Issue I worked on

see https://openmrs.atlassian.net/browse/SDK-

- [x] I have **added tests** to cover my changes. (If you refactored


sdk-commons/src/main/resources/openmrs-signing-key.asc
```diff
@@ -0,0 +1,55 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
ibacher, line 1: At the end of the day, I don't want to hard-code this because we need to be able to expire and rotate this key. I think for this use-case it would be best to either:

1. Rely on the Maven artifact we already need for downloading artifacts
2. Pulling from the key server (e.g., https://keyserver.ubuntu.com/pks/lookup?search=infrastructure%40openmrs.org&fingerprint=on&op=index)

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
```diff
@@ -63,6 +67,79 @@ public void downloadArtifact(Artifact artifact, File directory, boolean unpack)
...
+	}
+
+	public void verifyAlreadyDownloaded(Artifact artifact, File artifactFile, File directory) throws MojoExecutionException {
+		Artifact ascArtifact = makeAscArtifact(artifact);
+		File ascFile = new File(directory, ascArtifact.getDestFileName());
+
+		boolean signatureFetched = tryFetchSignature(ascArtifact, directory);
+
+		if (!signatureFetched || !ascFile.exists()) {
+			log.warn("Artifact {} is not signed (pre-backfill release). Skipping verification.", artifact);
```
ibacher: Remember, though, there are also artifacts we don't publish that the SDK may be managing, so it's not necessarily a "pre-backfill release". It would also be good to validate that the artifact at least is associated with the `org.openmrs` group as we'll never sign anything outside of that group.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/MavenEnvironment.java
```diff
@@ -22,6 +22,8 @@ public class MavenEnvironment {
 	private BuildPluginManager pluginManager;
 	private Wizard wizard;
 
+	private boolean verifySignatures = true;
```
ibacher: This doesn't seem useful, but also it violates the contract of what this class is about.

sdk-commons/src/main/resources/openmrs-signing-key.asc
wikumChamith: I  switched to using the key server.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/MavenEnvironment.java
wikumChamith: Dropped this

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
wikumChamith: Done

ibacher:
Now a design question: can we leverage that new Maven plugin to do the signature validation and just need to do that in one place or does this make more sense?

wikumChamith:
> Now a design question: can we leverage that new Maven plugin to do the signature validation and just need to do that in one place or does this make more sense?

Yeah, I think we have multiple ways to do this, and I'm not sure which one to choose :) We could make the classes in the plugin `public`, move them into a separate reusable artifact (though this feels like an overreach), or add a new plugin goal and invoke it from the SDK (this feels cleaner). What do you suggest?

ibacher:
I think plugin goal makes the most sense and it's pretty consistent with how the SDK already works (as a Frankenstein's monster of various plugin invocations with a bit of random code-padding).

wikumChamith:
> I think plugin goal makes the most sense and it's pretty consistent with how the SDK already works (as a Frankenstein's monster of various plugin invocations with a bit of random code-padding).

@ibacher PR to create the plugin: https://github.com/openmrs/openmrs-contrib-pgpverify-maven-plugin/pull/2

I'll update this PR to use it once that gets merged.

wikumChamith:
@ibacher, @dkayiwa, I updated this to use openmrs-pgpverify-maven-plugin.

wikumChamith:
@ibacher, it looks like we haven't signed reference metadata module artifacts: https://openmrs.jfrog.io/ui/repos/tree/General/modules/org/openmrs/module/referencemetadata-omod/2.13.0

dkayiwa (commented):
This isn't ready to merge yet, and the red CI is the blocker, though not for the reason your last comment points to.

With verification on by default, the signature check is failing `setup`, `deploy`, and `build-distro` in this PR's own CI, and it is failing on transient network errors rather than on bad signatures. The mechanism and a suggested fix are on the `verify-files` call in `ArtifactHelper`.

On the unsigned reference-metadata artifacts from your last comment: that isn't what is failing CI. Unsigned artifacts are tolerated correctly here (the run logs dozens of `no PGP signature (.asc) found` warnings that do not fail the build). Both jobs died on `504 Gateway Time-out` responses fetching `.asc` files from oss.sonatype.org: referencemetadata-omod and referenceapplication-omod for `setup`, and namephonetics/appui/htmlformentry/registrationapp/allergyapi for `deploy`. Signing more modules won't turn CI green; the transport-error handling has to change first.

Two smaller points are inline and don't block: whether unpacked artifacts should be verified too, and the missing opt-out. The transient-error handling is the one that matters for merge.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
```diff
@@ -63,6 +63,78 @@ public void downloadArtifact(Artifact artifact, File directory, boolean unpack)
...
+		executeMojo(
+				plugin(
+						groupId(SDKConstants.PGPVERIFY_PLUGIN_GROUP_ID),
+						artifactId(SDKConstants.PGPVERIFY_PLUGIN_ARTIFACT_ID),
+						version(SDKConstants.PGPVERIFY_PLUGIN_VERSION)
+				),
+				goal("verify-files"),
+				configuration(
+						verifyFilesArtifacts(openmrsArtifacts, directory),
+						element("failOnMissingSignature", "false")
```
dkayiwa: `failOnMissingSignature=false` only covers the case where the `.asc` genuinely does not exist (a 404 from every repo). It does not cover the case where the `.asc` lookup *fails* for some other reason, and that is what is turning this PR's CI red.

In the latest run, `deploy` (downgrade to referenceapplication:2.1) and `setup` both failed with `PGP signature verification failed for N artifact(s)`, but the cause was not a bad or missing signature. It was `could not resolve PGP signature: ... status code: 504, reason phrase: Gateway Time-out` while fetching `.asc` files from oss.sonatype.org. The plugin does this on purpose: in `AbstractVerifyMojo`, a resolution error is only downgraded to "missing" when every failure is `ArtifactNotFoundException`; a 5xx/transport error is re-thrown and fails the build regardless of `failOnMissingSignature`.

Because verification now runs by default over the project's full repository list (which includes oss.sonatype.org) on `setup`, `deploy`, and `build-distro`, any transient hiccup fetching a signature (even for an artifact that is not signed at all) breaks the user's command. If merged as-is, `setup -Ddistro=referenceapplication:2.x` and `build-distro` will fail intermittently for end users whenever a configured repo returns a 5xx/timeout on a `.asc` lookup. A green re-run would not clear the design problem: this turns "a mirror was briefly slow" into "the command failed".

Since you own the plugin, I would fix it there: treat an unresolvable `.asc` (transport/5xx) the same as a missing one when `failOnMissingSignature=false` (log and skip), and keep hard failures only for a signature that is present and cryptographically invalid. If you would rather stay fail-closed, resolve the `.asc` only from the canonical OpenMRS repo (with a retry) so a random flaky mirror cannot gate the build. I would default to the first option, since it matches what the README already promises (missing -> skip, invalid -> fail).

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
```diff
@@ -63,6 +63,78 @@ public void downloadArtifact(Artifact artifact, File directory, boolean unpack)
 							mavenEnvironment.getPluginManager()
 					)
 			);
+
+			if (!unpack) {
```
dkayiwa: Is it intended that verification only runs on the `copy` path and skips everything downloaded with `unpack=true`? The unpack callers include content packages (`ContentHelper`) and the frontend SPA zip (`DistributionBuilder`: `new Artifact(frontendArtifactId, distroVersion, distroGroupId, "zip")`), and those can be `org.openmrs.*` artifacts that `isOpenmrsArtifact` would otherwise match. The PR description says modules, WAR and OWAs are verified, so I want to check whether leaving unpacked artifacts unverified is a deliberate v1 boundary or an oversight. If deliberate, it is worth saying so in the description/README.

maven-plugin/src/main/resources/build-distro/README.md
```diff
@@ -18,3 +18,7 @@ To regenerate this distribution:
 mvn openmrs-sdk:build-distro -Ddistro=openmrs-distro.properties
 ```
 Add `-Dreset` to overwrite any files you have customised.
+
+Downloaded `org.openmrs` artifacts are verified against the OpenMRS GPG signing key.
+Artifacts without a signature are logged and skipped; an artifact whose signature is
+present but invalid fails the build.
```
dkayiwa: Non-blocking, on the opt-out: the PR description still tells users to disable this with `-DverifySignatures=false`, but that flag was dropped earlier in this PR, so a feature that is on by default and gates every download currently has no documented off switch. The plugin does expose `openmrs.pgpverify.skip` (property `openmrs.pgpverify.skip`); it would be worth confirming that propagates through the `executeMojo` invocation and documenting it here as the escape hatch, or updating the description if the intent is that it cannot be turned off.

wikumChamith:
@dkayiwa, @ibacher the build errors are now fixed.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
dkayiwa: I bumped this to 1.0.1 and traced it through, and it isn't closed yet. 1.0.1 adds a `signatureRepository` parameter that lets you pin `.asc` resolution to one repo, but this `verify-files` call doesn't set it, so signatures are still resolved from the full project repository list (the mirrors in `settings.xml`, including `oss.sonatype.org`). And the transport-error classification is unchanged: `isGenuinelyMissing` counts a failure as "missing" only when every cause is `ArtifactNotFoundException`, so a 504 or timeout on a `.asc` lookup is rethrown and fails the build even with `failOnMissingSignature=false` (that flag only covers the genuinely-absent case). The plugin even ships a `verify-transport-error` IT asserting this on purpose, so it's intended behavior in 1.0.1, not something the version bump relaxed.

So if merged as-is, a briefly slow or unreachable mirror contacted during a `.asc` lookup fails the user's `setup`/`build-distro`/`deploy` with `could not resolve PGP signature`, even when nothing is actually wrong with the signature. That's the same case I flagged before; CI is green only because no repo happened to time out on a `.asc` fetch during this run, not because the path changed.

Simplest fix now that 1.0.1 exists: set `signatureRepository` on this invocation to the canonical OpenMRS repo, since that's where `org.openmrs` `.asc` files live anyway, so a flaky third-party mirror can't gate the check. Concretely, add `element("signatureRepository", "https://mavenrepo.openmrs.org/public")` next to the `failOnMissingSignature` element. If you'd rather keep the full repo list, the alternative is to make the plugin treat an unresolvable `.asc` the same as a missing one when `failOnMissingSignature=false`. I'd go with setting `signatureRepository` here.

wikumChamith: Done!

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
```diff
@@ -63,6 +63,79 @@ public void downloadArtifact(Artifact artifact, File directory, boolean unpack)
...
+				plugin(
+						groupId(SDKConstants.PGPVERIFY_PLUGIN_GROUP_ID),
+						artifactId(SDKConstants.PGPVERIFY_PLUGIN_ARTIFACT_ID),
+						version(SDKConstants.PGPVERIFY_PLUGIN_VERSION)
+				),
+				goal("verify-files"),
+				configuration(
+						verifyFilesArtifacts(openmrsArtifacts, directory),
+						element("failOnMissingSignature", "false"),
+						element("signatureRepository", SDKConstants.OPENMRS_SIGNATURE_REPOSITORY)
```
dkayiwa: This needs one more fix before it can go in. Pinning the `.asc` lookup closes the mirror problem, but the other network leg, fetching the public key, is still unguarded.

With no `keyRings` passed, the plugin defaults `keyServer` to `https://keyserver.ubuntu.com` and fetches the signing key over a bare `HttpURLConnection`. That fetch never lands in the local repository (`KeyServerClient` keeps an in-memory map on a mojo instance Maven recreates per execution), and it does not go through Maven's resolver, so no mirror, no `settings.xml` proxy, no retry. Any failure surfaces in `SignatureVerifier.verify` as an `IOException`, gets wrapped as a `VerificationException`, and becomes a build error. `failOnMissingSignature=false` does not cover it.

I ran the real `openmrs-webapp-2.6.0.war` through `verify-files` with exactly the configuration this method builds. Key server reachable:

```
[INFO] OK   org.openmrs.web:openmrs-webapp:2.6.0 signed by CA12619FDE8CD6A93FAFE458A6F9608DCC73473F
[INFO] BUILD SUCCESS
```

Same command with the key server pointed at an unresolvable host:

```
[ERROR] org.openmrs.web:openmrs-webapp:2.6.0 - could not read artifact data for verification: keyserver.invalid
[ERROR] PGP signature verification failed for 1 artifact(s)
[INFO] BUILD FAILURE
```

So if merged as-is, `setup`, `deploy` and `build-distro` fail outright for anyone who cannot reach keyserver.ubuntu.com, and that is not a corner case. Every `openmrs-webapp` version the ITs install (2.0.2, 2.3.1, 2.5.9, 2.6.0, 2.6.7, 2.6.9, 2.7.0) now has a `.asc` in the OpenMRS repo, so every setup reaches the key server. Offline and low-bandwidth installs, and anyone whose proxy is configured only in `settings.xml`, get a failed command instead of a warning.

What I would do here: hand the plugin the bot public key as a `keyRings` entry and set `keyServer` explicitly to the Ubuntu server. `PublicKeySource` checks the ring first and only calls out for a key it does not have, so the normal path stops touching the network while a rotated key still resolves through the server. Two details worth knowing before you wire it up: supplying `keyRings` suppresses the default key server, so it has to be set explicitly or you lose the fallback entirely, and `Locations.open` takes a file path or an http(s) URL but not a classpath resource, so a key bundled in the jar needs extracting to a temp file first. I confirmed the shape works: same artifact, same unresolvable key server, with `<keyRings>` pointing at the exported bot key it verifies and the build passes.

This does not reopen the hard-coding discussion from earlier in the PR. The trust decision is the fingerprint pin, which lives in the plugin either way; the ring is only a local copy of public key material. The alternative is to make a key-source failure degrade to a warning in the plugin the way a missing signature already does, but that silently switches verification off whenever the key server is down, so I would go with the ring.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/SDKConstants.java
```diff
@@ -19,6 +19,11 @@ public class SDKConstants {
     public static final String DEPENDENCY_PLUGIN_GROUP_ID = "org.apache.maven.plugins";
     public static final String DEPENDENCY_PLUGIN_ARTIFACT_ID = "maven-dependency-plugin";
     public static final String DEPENDENCY_PLUGIN_VERSION = "3.2.0";
+    // pgpverify plugin
+    public static final String PGPVERIFY_PLUGIN_GROUP_ID = "org.openmrs.maven.plugins";
+    public static final String PGPVERIFY_PLUGIN_ARTIFACT_ID = "openmrs-pgpverify-maven-plugin";
+    public static final String PGPVERIFY_PLUGIN_VERSION = "1.0.1";
```
dkayiwa, line 25: With no `keysFile` or `groups` passed to the plugin, the trusted fingerprint set is whatever `AbstractVerifyMojo.OPENMRS_BOT_KEY` was compiled as in 1.0.1, so rotating the signing key means cutting a new plugin release, bumping this constant, and cutting a new SDK release. Until users upgrade, anything signed with the new key fails with `signed by untrusted key ... (allowed: [CA12619FDE8CD6A93FAFE458A6F9608DCC73473F])`, which is a hard build failure, not a warning.

Is that the intended rotation story? Moving to the key server solved where the key material comes from, but the trust anchor is still baked into a pinned plugin version, which is the part of @ibacher's original point that I think is still open. The plugin's `keysFile` parameter looks like it exists for exactly this, though pointing it at a URL introduces another hard network dependency, so I do not have a clean recommendation. Mainly want to be sure it is a deliberate trade rather than something that slipped.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/SDKConstants.java
```diff
@@ -19,6 +19,11 @@ public class SDKConstants {
...
+public static final String OPENMRS_SIGNATURE_REPOSITORY = "https://mavenrepo.openmrs.org/public";
```
dkayiwa: This line lost its indentation.

```suggestion
    public static final String OPENMRS_SIGNATURE_REPOSITORY = "https://mavenrepo.openmrs.org/public";
```

sdk-commons/src/test/java/org/openmrs/maven/plugins/utility/ArtifactHelperTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.Test;
+import org.openmrs.maven.plugins.model.Artifact;
+
+import java.io.File;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.nullValue;
+
+public class ArtifactHelperTest {
```
dkayiwa, line 13: Nothing in the suite fails if verification stops happening. These three cover the `Xpp3Dom` mapping, which is worth having, but they would all still pass if the `verifySignatures` calls were deleted from both call sites, or if the plugin skipped every artifact at runtime.

Not blocking from my side, but for a security feature I would want one test that pins the runtime effect. The cheapest honest one I can see is an IT next to `ArtifactHelperIT` that installs an `org.openmrs` artifact whose on-disk copy has been corrupted and asserts the build fails with the verification error, plus an assertion on a normal run that the log carries `PGP verification checked N whitelisted artifact(s)` with N greater than zero. As it stands this can regress to a no-op and CI stays green.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
dkayiwa: The `!unpack` branch never runs outside the integration tests. Every production call site passes `unpack=true`: `ContentHelper` at 55, 124 and 160, `DistributionBuilder` at 74 and 148, `DistroHelper` at 400. The only `false` callers in the tree are in `ArtifactHelperIT`. So the verification added to `downloadArtifact` is not checking anything in a real run, and all the real coverage comes from `ModuleInstaller.prepareModules` on the `copy` path (modules, WAR, OWAs).

That also makes the line this PR adds to the build-distro README inaccurate. "Downloaded `org.openmrs` artifacts are verified against the OpenMRS GPG signing key" reads as covering the content packages and the frontend SPA zip, and those are `org.openmrs.*` artifacts going through the unpack path unverified. For a security feature that gap matters more than usual, because someone reads that sentence and believes their content package was checked.

Two ways out. Either verify the unpack path as well, which needs a copy-then-verify-then-unpack since the archive is gone once `unpack` has run, or drop the unreachable branch and narrow the README to modules, WAR and OWAs. I would take the first if content packages are inside the threat model you are aiming at, otherwise the second, but leaving a branch that only the ITs reach plus a README that overstates the coverage is the one option I would not ship.

wikumChamith:
@dkayiwa I updated the PR.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/ModuleInstaller.java
```diff
@@ -165,5 +165,9 @@ private void prepareModules(Artifact[] artifacts, String outputDir, String goal,
                         mavenEnvironment.getPluginManager()
                 )
         );
+
+        if (GOAL_COPY.equals(goal)) {
+            mavenEnvironment.getArtifactHelper().verifySignatures(Arrays.asList(artifacts), new File(outputDir));
```
dkayiwa: The verification here only fires for callers that route through `ModuleInstaller`, and the single-module deploy path does not, so `deploy` still installs modules unchecked. `Deploy.deployModule(String, String, String, Server)` builds its own `artifactItems` and calls `executeMojoPlugin` (Deploy.java:382), which runs `dependency:copy` directly and never reaches this method. That is the path behind `openmrs-sdk:deploy -DartifactId=... -Dversion=...` (Deploy.java:161), the interactive "Module" choice (177), deploy from a module project directory (145), and the `owa-omod` install inside `deployOwa` (252).

If merged as-is, `openmrs-sdk:deploy -DartifactId=xforms -Dversion=5.0.0` resolves that module from the configured repositories, writes it into the server's `modules` directory, and prints nothing about signatures, so nobody notices the check did not run. In that same OWA command the zip itself is verified, because it goes through `installModule`, while the `owa-omod` pulled in ahead of it is not, so there is no way to tell from the outside which of your deploys were covered.

`Deploy` already holds a `moduleInstaller` (`AbstractTask`:108, created at 204 from the same `mavenEnvironment`), and `installModule` runs the same `dependency:copy` over the same `artifact.toElement(outputDir)` element, just with the verification after it. So at Deploy.java:382 this is a drop-in, and the local `artifactItems` list then becomes unused:

```java
moduleInstaller.installModule(artifact, modules.getPath());
```

I would leave the other `executeMojoPlugin` caller alone. `deployOpenmrsWar` at 354 deliberately installs the WAR built in the current directory, and checking that against the released signature would be wrong. If you would rather keep `deploy` outside this PR, that is a reasonable call, but then please say so in the description, since half of the same command already verifies and that is invisible to users.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
```diff
@@ -65,4 +69,109 @@ public void downloadArtifact(Artifact artifact, File directory, boolean unpack)
...
+		item.add(element("groupId", artifact.getGroupId()));
+		item.add(element("artifactId", artifact.getArtifactId()));
+		item.add(element("version", artifact.getVersion()));
+		item.add(element("type", artifact.getType()));
+		if (artifact.getClassifier() != null) {
+			item.add(element("classifier", artifact.getClassifier()));
+		}
+		return element("artifact", item.toArray(new MojoExecutor.Element[0]));
+	}
+}
```
dkayiwa: This file and the new `ArtifactHelperTest.java` both end without a trailing newline, which is why git prints `\ No newline at end of file` on them. Trivial, but it makes every later diff of these two files a line noisier than it needs to be.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
dkayiwa: Going back over the call sites I listed above, I mislabeled one: `DistributionBuilder:74` is not the SPA zip, it is the distro artifact itself, downloaded and unpacked to read `openmrs-distro.properties`. `DistroHelper.downloadDistro` fetches it as well, also unverified. So the new README sentence, "Content packages and the frontend SPA are not verified", is still missing the item with the most leverage, since that file is what decides which modules and versions get installed in the first place.

Verifying it would not close this either: `referenceapplication-package-2.13.0.jar` has no `.asc` in mavenrepo.openmrs.org (404), so a check there would only ever warn. What I would do is name the distro artifact in that sentence. One thing worth knowing while you word it: distro properties can set `omod.<name>.groupId` to anything, and groups outside `org.openmrs` are skipped entirely, so a substituted distro definition sidesteps the check rather than tripping it. Not blocking, but for a security feature I would rather the shipped README understate the coverage than overstate it.

maven-plugin/src/main/resources/build-distro/README.md
dkayiwa: `-Dopenmrs.pgpverify.skip=true` does reach the plugin, so the escape hatch exists. The 1.0.1 descriptor declares `<skip implementation="boolean" default-value="false">${openmrs.pgpverify.skip}</skip>`, and mojo-executor merges the descriptor's configuration into the execution (`MojoExecutor.java:131`), so that expression survives for any parameter `verifySignatures` does not set itself. The three it does set, `keyServer`, `signatureRepository` and `failOnMissingSignature`, win over their properties and cannot be overridden from the command line, so `skip` is the only switch a user has.

Still worth a line in the README, because the check hard-fails. I ran `verify-files` with exactly the configuration this method builds against a release-versioned artifact whose bytes did not match the published signature, and it exits with `signature is INVALID` and `BUILD FAILURE`. Since `dependency:copy` takes the local repository copy for a release version, a stale or locally rebuilt jar in `~/.m2` would trip that during `setup`, and right now nothing tells the user how to get past it while they work out why.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/ModuleInstaller.java
wikumChamith: Done!

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
wikumChamith: Done !

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/ModuleInstaller.java
```diff
@@ -165,5 +165,9 @@ private void prepareModules(Artifact[] artifacts, String outputDir, String goal,
...
+        }
```
dkayiwa, line 179: This runs after `dependency:copy` has already written the artifact into its final location, and nothing removes it when the check fails. `setup` gets away with that because `Setup.executeTask` deletes the whole server directory when anything throws (Setup.java:662), but `deploy` has no equivalent cleanup.

If merged as-is, `openmrs-sdk:deploy -DartifactId=xforms -Dversion=5.0.0` deletes the currently installed copy in `deleteModuleFromServer`, copies the replacement into `<server>/modules`, then fails here. The rejected `.omod` is still on disk, and the next `openmrs-sdk:run` starts OpenMRS with it: `Run.executeTask` does no verification, and since the SDK never writes `module_list_to_load` into the runtime properties, `ModuleFactory.loadModules()` just lists the modules folder and loads every file in it. So the check gates the command but not the server. For the case this feature exists for, a well-formed module substituted by a compromised mirror, the artifact that failed verification is the one that ends up running, and by then nothing is warning about it.

I reproduced the residue using this method's order, `dependency:copy` then `verify-files`, against a release jar whose local repository copy I had corrupted:

```
[INFO] Copying xforms-omod-5.0.0.jar to .../modules/xforms-5.0.0.omod
[ERROR] org.openmrs.module:xforms-omod:5.0.0 - signature is INVALID
[INFO] BUILD FAILURE
$ ls modules/
xforms-5.0.0.omod
```

Deleting what was just copied is enough to close it, and every class it needs is already imported in this file:

```suggestion
        if (GOAL_COPY.equals(goal)) {
            try {
                mavenEnvironment.getArtifactHelper().verifySignatures(Arrays.asList(artifacts), new File(outputDir));
            }
            catch (MojoExecutionException e) {
                for (Artifact artifact : artifacts) {
                    FileUtils.deleteQuietly(new File(outputDir, artifact.getDestFileName()));
                }
                throw e;
            }
        }
```

The sturdier shape is to copy into a temp directory, verify there, and move into `outputDir` only on success, so untrusted bytes never reach the server directory at all. I would do the delete first, since it closes the hole without changing how this method works.

sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/ArtifactHelper.java
dkayiwa: `ArtifactHelperTest.java` still ends without a trailing newline, so the diff keeps printing `\ No newline at end of file` for that one.

sdk-commons/src/test/java/org/openmrs/maven/plugins/utility/ArtifactHelperTest.java
dkayiwa: What is still not pinned is the wiring rather than the check. Both new ITs call `verifySignatures` directly, so deleting the call in `ModuleInstaller.prepareModules`, now the only production call site, leaves the whole suite green. One line in an existing end-to-end IT covers it: `SetupIT` already runs a real `setup` from the outer JVM, so `verifier.verifyTextInLog("PGP verification checked")` alongside the `assertSuccess()` there starts failing as soon as the call goes away. Not blocking.

maven-plugin/src/main/java/org/openmrs/maven/plugins/utility/ModuleInstaller.java
wikumChamith: Done!

ibacher (approved):
But I agree that it would be ideal to have the end-to-end test @dkayiwa mentions [here](https://github.com/openmrs/openmrs-sdk/pull/373/changes#r3673745975).

wikumChamith:
> But I agree that it would be ideal to have the end-to-end test @dkayiwa mentions [here](https://github.com/openmrs/openmrs-sdk/pull/373/changes#r3673745975).

Done!

maven-plugin/src/main/java/org/openmrs/maven/plugins/Deploy.java
```diff
@@ -369,17 +369,15 @@ public void deployOpenmrsWar(Server server, Artifact artifact) throws MojoExecut
...
 
 		File modules = new File(server.getServerDirectory(), SDKConstants.OPENMRS_SERVER_MODULES);
 		modules.mkdirs();
-		artifactItems.add(artifact.toElement(modules.getPath()));
 
 		boolean moduleRemoved = deleteModuleFromServer(artifact, modules, server);
 
 		if (moduleRemoved) {
-			executeMojoPlugin(artifactItems);
+			moduleInstaller.installModule(artifact, modules.getPath());
```
dkayiwa, line 381: Was leaving `deployOpenmrsWar` on the old path deliberate? This call covers the module side of `deploy`, but `deployOpenmrsWar` still goes through `executeMojoPlugin` at line 354, which makes it the one remaining `dependency:copy` in this class that installs an artifact without the check.

There is a fair case for that. It is the "deploy the war from the current directory" path, and verifying a locally built war against the published signature would fail. But the plugin already skips SNAPSHOTs, so that only bites on a release-versioned local build, and in the other direction an openmrs-core clone sitting on a release tag that has not been built locally makes `dependency:copy` fetch the published war and install it unchecked.

Either answer works for me, I would just rather it not be accidental. If it is deliberate, a short comment at 354 saves the next person from "fixing" it; if not, the same `moduleInstaller.installModule(artifact, server.getServerDirectory().getPath())` swap fits there too.

integration-tests/src/test/java/org/openmrs/maven/plugins/SetupIT.java
```diff
@@ -46,6 +46,7 @@ public void setup_shouldInstallRefapp2_3_1() throws Exception{
         executeTask("setup");
 
         assertSuccess();
+        verifier.verifyTextInLog("PGP verification checked");
```
dkayiwa, line 49: This pins `setup`, and nothing pins `deploy`. `Deploy.deployModule` moved from `executeMojoPlugin` to `moduleInstaller.installModule` in this PR, and if that move were reverted, `DeployIT.deploy_shouldInstallModule` would stay green, because all it asserts is that the `.omod` landed.

The same one-liner after `assertSuccess()` there covers it. That test deploys `org.openmrs.module:owa-omod:1.4`, a release version, so the plugin runs and logs the line, and `setupTestServer` deletes `log.txt` before `executeTask("deploy")` runs, so the match cannot come from the setup phase.

Not blocking, it is just the cheapest way to keep the path this PR rewired from quietly reverting to an unverified copy.

maven-plugin/src/main/resources/build-distro/README.md
dkayiwa: The new paragraph does not mention the off switch, so a check that hard-fails still leaves the user nothing to get past it with. One more sentence after "are not verified." would do it:

Pass `-Dopenmrs.pgpverify.skip=true` to skip the check, for example when a locally rebuilt copy of a release artifact in `~/.m2` no longer matches its published signature.

